Issue #23, opened 2020-06-23
Improve compatibility with third party video players


thirtythreeforty:
**Describe the bug**
Currently many video players cannot play video coming out of the RTSP server (see trouble users are having in #19 for example).  This is probably because their receive buffers are too small for a 4K keyframe, or else the OS is dropping UDP packets because each keyframe is a lot of traffic all at once.

**To Reproduce**
Open Neolink's RTSP stream from any Reolink camera in VLC.  No video.  In other players, at best it's garbled.

**Expected behavior**
The video should play, without any monkeying with receive buffers, protocols, or other technical details.

Ideas: "smooth out" the incoming key frames that cause so many problems by rate-limiting them just a little bit.

twistedddx:
If the file format you are seeing is the same as the files the official client writes to files it may explain your 3rd party player compatibility also.
You are somewhat depending on luck for the decoder to recognize the h264 video data hidden between other unrecognized data.
Eg the video decoder is having to deal with BC headers and likely also has a bunch of audio data at the end of each frame.

I say BC headers and file format but it is just as likely the format was created by Hisilicon.

I personally found that any VLC higher than 2.2.8 on Windows can not correctly select its raw h264 decoder even if manually set. 
So modern VLC is a poor choice to try to force playback of h264 BC files in their non standard container.
Other FFmpeg based decoders do a better job in that they do not suffer from this bug, eg LAV filters can often play the video portion of the files like you want VLC to do.

FFmpeg's AAC decoder is currently unable to luck auto recognize the audio data within BC format.

thirtythreeforty:
Interesting, thank you for the [link to the file format](https://www.wasteofcash.com/BCConvert/BC_fileformat.txt) in #4. I had run across your tool, but not the text document.

The Gstreamer H.265 frame finder embedded in Neolink successfully recognizes the H.265 stream from the raw `binaryData` stream passed from the cameras. There's no guarantee made that this is actually raw H.265, however, I had just assumed... I will reopen some of my pcaps and see if the contents of binaryData match the header format you describe.

thirtythreeforty:
I'm going to punch something if there's another stateful parser layer...

thirtythreeforty:
This is completely fixed in my testing with VLC. @QuantumEntangledAndy added PTS and DTS timestamps to the stream and it's completely fluid.

PR #24, opened 2020-06-30
Add H264 and subStream

### LINKED ISSUE #18: Substream proxying ?
Is it possible to pull the substream from camera into neolink rather than just the 4k stream ?

Reason behind this is for a few reasons.
1. I record 24x7, I'd like to optimise the storage by recording a 24x7 low/medium stream, and then on motion detection, switch to recording the HD stream.

2. Processing 4k video streams for real time object detection is painful

Nice to have: Some method of setting the substream output size stream would be useful, unsure if gstreamer can do this natively, or would require ffmpeg

### LINKED ISSUE #19: Support for H.264 video (E1, D400, etc)
I have the Reolink E1 camera and neolink definitely seems to be able to connect and talk to it, judging by the debug output. When connecting with RTSP to it using VLC, I get a streaming error in VLC and one log-row in neolink:
```[2020-06-20T19:30:03Z DEBUG neolink::gst] RTSP: media was configured```

Attached is a log from the execution with the latest Windows CI-build (https://github.com/thirtythreeforty/neolink/commit/9a0eb567dd99edd2147c176eb9948e0717f8d2b9)
[neolink-execution-9a0eb56.log](https://github.com/thirtythreeforty/neolink/files/4808470/neolink-execution-9a0eb56.log)

Can you find anything in the logs to indicate what can be done to support the E1 camera?
Thanks

### PR DESCRIPTION
**Pull request by QuantumEntangledAndy**
This PR adds H264 streaming and allows specifying which stream to use (HD or SD).

It adds two config variables to the `[[cameras]]`

```
format = ["h264" | "h265"]
stream = ["mainStream" | "subStream"]
```

The defaults are set to "h265" and "mainStream" so that no changes are needed to current config files.

fixes #19
fixes #18

I have tested both of these features on my E1 camera.

I will note that the E1 camera is streaming raw H264 frames so that there are no timestamps. (DTS and PTS are always zero). This means that certain players (e.g. VLC) that drop frames with duplicate PTS will only show the first frame.

I have also added the ability to arbitrarily change the gst launch string. This was mostly used for debugging (since recompiling on a rasppi takes a long time just to change one string) but this may be useful if someone wants to change the streaming format so I have left the ability in. This can be done by setting `format = "! gststuf..."` in the config.

### CODE REVIEW

**Comment by QuantumEntangledAndy** [comment]
I think I have the issue with playing in VLC sorted now by getting gstreamer to add the time stamping information. I shall test it a bit more and if it works I will add it to this PR.

**Comment by QuantumEntangledAndy** [comment]
Ok so I get gstreamer to add time stamps to the stream and suddenly it is playable in VLC, and ffmpeg without any special changes and options. This may help more generally with #23

**Comment by QuantumEntangledAndy** [comment]
p.s. this is my first time with rust, (usually a python fan) so I suspect some of my code can be improved

**Comment by thirtythreeforty** [comment]
First, huge thank you for testing with your camera and for implementing this. Awesome.

> This PR adds H264 streaming and allows specifying which stream to use (HD or SD).
> 
> It adds two config variables to the `[[cameras]]`
> 
> ```
> format = ["h264" | "h265"]
> stream = ["mainStream" | "subStream"]
> ```
> 
> The defaults are set to "h265" and "mainStream" so that no changes are needed to current config files.

A couple things here. First, manually specifying the format is probably okay as a first pass, but I would like to go back and autodetect the format so that users do not have to know or care what format their camera uses. This can be done at the gstreamer layer with typefind, or by examining the camera's metadata that it sends and that we mostly currently ignore.

I have wanted to add substreams for a little while now, it's really nice to know that this "just works" from a streaming perspective.

The other wrinkle is that substream does not "just work" on my B800 - I suspect but have not verified that the substream is H.264; let me retest now that your other changes work.

> I will note that the E1 camera is streaming raw H264 frames so that there are no timestamps. (DTS and PTS are always zero). This means that certain players (e.g. VLC) that drop frames with duplicate PTS will only show the first frame.

You are the new video codec guru for Neolink, here's your badge. Is there a way you figured this out or just accumulated knowledge?

> I have also added the ability to arbitrarily change the gst launch string. This was mostly used for debugging (since recompiling on a rasppi takes a long time just to change one string) but this may be useful if someone wants to change the streaming format so I have left the ability in. This can be done by setting `format = "! gststuf..."` in the config.

Cool. I think I would prefer leaving this undocumented, because it'll only be useful for debugging, and perhaps moving it to a different configuration item, as in `gst_launch = "! your ! pipeline"`.

I have a couple comments on the code but in general it looks good - stand by for them. Again thanks, these features are desirable and this is likely to be merged in some capacity!

File: README.md [hunk 1]
```diff
@@ -63,6 +63,10 @@ Copy and modify the `sample_config.toml` to specify the address, username, and p
 Each `[[cameras]]` block creates a new camera; the `name` determines the RTSP path you should connect your client to.
 Currently Neolink cannot auto-detect cameras like the official clients do; you must specify their IP addresses directly.
 
+By default the h265 streaming format is used. Some cameras, for example E1, provide h264 streams, to use these you must specify `format = "h264"` in the `[[cameras]]` config.
+
+By default the HD stream is used. To swap to the SD stream add `stream = "subStream"` to the `[[cameras]]` config. You may also need to use `h264` format while using the SD stream.
```
thirtythreeforty, line 68: I'd like to have Neolink provide both the main stream and the sub stream for clients to connect to. I think this would look like `rtsp://your_server:8554/your_camera/subStream` and the corresponding `your_camera/mainStream` but I am open to ideas how this should work. Then you wouldn't need this configuration item, and you'd just connect to whatever you want in the client.

Feedback welcome, what do you think?

src/bc_protocol.rs
```diff
@@ -242,7 +242,7 @@ impl BcCamera {
         Ok(())
     }
 
-    pub fn start_video(&self, data_out: &mut dyn Write) -> Result<Never> {
+    pub fn start_video(&self, data_out: &mut dyn Write, stream: &String) -> Result<Never> {
```
thirtythreeforty, line 245: This should be `&str`, a reference to a string slice.  Can you rename it to `stream_name`? (Nitpick I know)

src/main.rs
```diff
@@ -43,7 +43,19 @@ fn main() -> Result<(), Error> {
...
+                    format = "! h265parse ! rtph265pay name=pay0";
+                } else if format == "h264" {
+                    format = "! h264parse ! rtph264pay name=pay0";
+                } else {
+                    if format.trim().starts_with("!") {
+                        warn!("Using custom format: {}.", format);
+                    } else {
+                        error!("Unrecognised format: {}.", format);
+                    }
+                }
```
thirtythreeforty, line 57: Picking gstreamer launch arguments should be done in the `gst` module, not here, to keep the gstreamer wrangling encapsulated.

Create a new enum:
```rust
enum StreamFormat {
    H264,
    H265,
    Custom(String),
}
```
in the gst module and add it as an argument to the `add_stream` function. In that function, create a `match` block that generates the right gstreamer arguments. Then, here in main, write a `match` block that transforms this configuration string into the appropriate variant of that enum.

thirtythreeforty: (It will be slightly more lines of code my way, but it will also keep the API nice and clean.)

thirtythreeforty:
> Ok so I get gstreamer to add time stamps to the stream and suddenly it is playable in VLC, and ffmpeg without any special changes and options. This may help more generally with #23

You have no idea how glad I am about this by the way, these pieces of software were making me very angry.

QuantumEntangledAndy:
> You are the new video codec guru for Neolink, here's your badge. Is there a way you figured this out or just accumulated knowledge?

About this it is really not guru like and is all monkeying. The process was chaotic at best:

- I streamed it in ffmpeg and it was always complaining about the lack of DTS and PTS.
- That was followed by lots of googling, which told me that some IP camera stream raw H.264 without the time information. This was usually on websites where people were unable to get it to work on VLC and were asking for help. Hmm sounds familiar...
- I then tested in ffmpeg by streaming through ffmpeg and adding PTS information with options like `-vf="PTS=N/Frame_rate/Baserate"` which made it work.
- Then back to google to see how to get gstreamer to add the right information, where I learned that there should be some Picture Timing SEI attached to the frames that includes timing formation. Finally this `https://stackoverflow.com/questions/61769447/feeding-video-to-gstreamer-from-c-how-to-get-timing-to-work` told me how to add timing to a frame in gstreamer `appsrc` by doing `"do-timestamp", TRUE`.

The solution I am using is to have gstreamer add the time stamp info from the clock, but this is not as accurate (but it works). If there is time stamp info somewhere in bc format it would maybe be best to add that to the bitstream Picture Timing SEI data.

Anyways I shall work on your changes and get back to you.

README.md
QuantumEntangledAndy: That sounds reasonable but can we make `rtsp://your_server:8554/your_camera` still point to `rtsp://your_server:8554/your_camera/mainStream` as well just so that people won't need to change their scripts

thirtythreeforty:
Yes - that's a good point - definitely make it backwards compatible.

src/bc_protocol.rs
QuantumEntangledAndy: This should be done now

src/main.rs
QuantumEntangledAndy: Ok moved also added validation to the config toml file so it will error if they don't write a valid input e.g. h264. Learned lost of rust today :)

QuantumEntangledAndy:
So I have added sub mounts and explicit port specification in the config

README.md
QuantumEntangledAndy: I may have this working just need to thoroughly test it

QuantumEntangledAndy:
Ok so I have just tested these features on my raspberry pi with the E1 all seems to be working, can access all three mount points `name`, `name/mainStream` and `name/subStream` with `name` being the same as `name/mainStream`. Please have a look at how I did it as it is maybe not the most efficient method.

QuantumEntangledAndy:
I have added the option `both` to the `stream = ` in the config, when it detects this it creates the three mounts `name`, `name/mainStream` and `name/subStream` however you can specify `stream = "subStream"` in which case it will not create the three mounts and will only create the mount you specify (main or sub) at `name`. I like this method, as the user can choose not to stream both if they want to.

QuantumEntangledAndy:
Next up is tackling `typefind` but that may be beyond my rust skills haha

src/main.rs
```diff
@@ -35,20 +41,63 @@ fn main() -> Result<(), Error> {
...
+    match config.validate() {
+        Ok(_) => (),
+        Err(e) => return Err(Error::ValidationError(e)),
+    };
+    for camera in &config.cameras {
+        match camera.validate() {
+            Ok(_) => (),
+            Err(e) => return Err(Error::ValidationError(e)),
+        };
+    }
```
thirtythreeforty, line 55: Hm - can validate() call the sub-validators automatically?

src/main.rs
```diff
@@ -35,20 +41,63 @@ fn main() -> Result<(), Error> {
     );
 
     let opt = Opt::from_args();
-    let config: Config = toml::from_str(&fs::read_to_string(opt.config)?)?;
+    let mut config: Config = toml::from_str(&fs::read_to_string(opt.config)?)?;
+
+    match config.validate() {
+        Ok(_) => (),
+        Err(e) => return Err(Error::ValidationError(e)),
+    };
```
thirtythreeforty, line 49: A handy Rust-ism for single-variant match:

```rust
if let Err(e) = config.validate() {
    // ...
}
```

src/main.rs
```diff
@@ -35,20 +41,63 @@ fn main() -> Result<(), Error> {
...
+
+            main_camera_config.stream = "mainStream".to_string();
+            main_camera_config.name = format!("{}/{}", main_camera_config.name, main_camera_config.stream);
+            new_cam_configs.push(main_camera_config);
+
+            sub_camera_config.stream = "subStream".to_string();
+            sub_camera_config.name = format!("{}/{}", sub_camera_config.name, sub_camera_config.stream);
+            sub_camera_config.format = "h264".to_string(); // Assuming always H264 on subStream: TODO: Autodetect
+            new_cam_configs.push(sub_camera_config);
+        }
```
thirtythreeforty, line 79: This approach will cause the main stream to stream from the camera twice, once for each RTSP path `camera/` and `camera/mainStream`. A better approach would be to call `add_stream()` once for each stream, possibly giving it more than one path. The "trick" here would be that you can call `crossbeam::scope::spawn` as many times as you like, not just once per loop.

If you want me to do this particular change, I can show you what I mean.

QuantumEntangledAndy: This means we would need to supply more than one path in the config and the loop over `mounts.add_factory(&format!("/{}", name), &factory);` with each path right? Is there a way to add a path array to the config structure without it being avaliable via toml I wonder... Hmm time to read more docs.

QuantumEntangledAndy: Aha! this `#[serde(skip)]` :)

src/main.rs
QuantumEntangledAndy: I did try first but it didn't validate it. Perhaps because it is inside a vector and it is not smart enough to test if the contents are validate-able. Maybe there's an option in the docs though.. I shall read into it more.

thirtythreeforty: Honestly don't worry about it for this PR. We can clean up if desired. Let's land the functionality.

src/main.rs
thirtythreeforty: Could use skip, although you don't need to change the config structure to get the code to do what you want here, I don't think. I'll write a patch on your branch if you don't mind, and go ahead and merge as it is now.

QuantumEntangledAndy: I see well it should be ok to merge then. I did try what I suggested with multiple paths and looping `mounts.add_factory` but it fails to setup the mounts. The docs at [https://gstreamer.freedesktop.org/documentation/gst-rtsp-server/rtsp-mount-points.html?gi-language=c#GstRTSPMountPoints] seems to suggest that once called the factory should not be used again and that calling add_factory also removes any previous mounts. Perhaps I am going about this the wrong way.

src/main.rs
QuantumEntangledAndy: I have it sorted out anyways here just add `#[validate]`

src/main.rs
QuantumEntangledAndy: So I have just updated the readme and validation and will now stop, as you say lets add the functionality and maybe make another PR for improvements when time permits. Thanks!

hh1209:
Huge thanks to @QuantumEntangledAndy for the development.

I would like to add that I have tested streaming both the mainStream and the subStream on my B800 and D800 cameras, and both streams worked with the three different paths. I could access both streams in VLC no trouble.

Getting the subStream to work in Blue Iris was a bit wonky... When you select rtsp protocol BI won't let you add a path for the mainstream, and setting the subStream path doesn't work either.

What I did to get both streams running in BI was to set the protocol to http and leave the URL 127.0.0.1:8554, then set the mainstream path to /camera_name/mainStream and the subStream path to /camera_name/subStream. Then subStream functionality worked for all 6 cams.

And my ram usage is down to a third of what it was!

src/main.rs
thirtythreeforty: Great! `mounts.add_factory` should be ok to call more than once, because the docs are talking about consuming the pointer you pass in, not the underlying Factory, which can be shared. This is transparent in the Rust bindings. (I *think*. I am not a Gstreamer expert.)

PR #25, opened 2020-07-03
Security: TLS and authentication

### LINKED ISSUE #5: Support RTSP security
Allow configuring a username/password for the served RTSP stream.  Also allow specifying a TLS certificate.

QuantumEntangledAndy:
This PR adds both TLS and basic authentication.

- The TLS controlled by the configuration options.

  - Adding `certificate = "PATHTOCERT"` will enable TLS, without client auth.
  - Also adding `tls_client_auth = "none|requested|required"` will require that the client provide a certificate signed by the server's CA.
  - When TLS is enabled standard `rtsp` is disabled and all connections must be through `rtsps`

- The basic auth is controlled by the options
```
username = "someuser"
password = "somepass"
```
If both user and pass is not present it will not enable basic auth.

Currently this PR enables one user to access all content or none. I expect to add to more fine grained control later.

As before my rust is basic, sorry about that.

fixes #5

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,10 @@ By default the HD stream is mounted at `name` and at `name/mainStream` and the S
 By default Neolink serves on all IP addresses on port 8554.
 You can modify this by changing the `bind` and the `bind_port` parameter.
 
+Use can enable `rtsps` (TLS) by adding a `certificate = "/path/to/pem"` to the config. This pem should contain by the certificate and the key used for the server. If TLS is enabled all connections must use `rtsps`. Use can also control client side TLS with the config option `tls_client_auth = "none|request|require"`, in this case the client should present a certificate signed by the server's CA, this is disabled by default.
+
+Use can enable basic authentication by adding the configuration option `username = "someone"` and `password = "somepass"`. Both username and password must be set for this to work.
+
 You can change the Neolink log level by setting the `RUST_LOG` environment variable (not in the configuration file) to one of `error`, `warn`, `info`, `debug`, or `trace`:
 
 ```
--- a/README.md
+++ b/README.md
@@ -112,7 +112,9 @@ Anywhere a username is accepted it can take any username or one of the following
 - `anyone` means any user with a valid user/pass
 - `anonymous` means no user/pass required
 
-The default `permitted_users` list is `[ "anonymous" ]` meaning no authentication required. If you want to use authenitcation then after adding your users through `[[users]]` please ensure you update the `permittted_users` for each camera.
+The default `permitted_users` list is:
+- `[ "anonymous"]` if no `[[users]]` were given in the config meaning no authentication required to connect.
+- `[ "anyone" ]` if `[[users]]` were provided meaning any authourised users can connect.
 
 You can change the Neolink log level by setting the `RUST_LOG` environment variable (not in the configuration file) to one of `error`, `warn`, `info`, `debug`, or `trace`:
 
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Currently Neolink cannot auto-detect cameras like the official clients do; you m
 
 By default the h265 streaming format is used. Some cameras, for example E1, provide h264 streams, to use these you must specify `format = "h264"` in the `[[cameras]]` config.
 
-By default the HD stream is used. To swap to the SD stream add `stream = "subStream"` to the `[[cameras]]` config. You may also need to use `h264` format while using the SD stream.
+By default the HD stream is mounted at `name` and at `name/mainStream` and the SD stream is mounted at `name/subStream`. You can specify mounting only one by adding `stream = "subStream"` to the `[[cameras]]` config. You may also need to use `h264` format while using the SD stream.
 
 By default Neolink serves on all IP addresses on port 8554.
 You can modify this by changing the `bind` and the `bind_port` parameter.
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ You may also need to use `h264` format while using the SD stream.
 By default Neolink serves on all IP addresses on port 8554.
 You can modify this by changing the `bind` and the `bind_port` parameter.
 
-Use can enable `rtsps` (TLS) by adding a `certificate = "/path/to/pem"` to the config. This pem should contain by the certificate and the key used for the server. If TLS is enabled all connections must use `rtsps`. Use can also control client side TLS with the config option `tls_client_auth = "none|request|require"`, in this case the client should present a certificate signed by the server's CA, this is disabled by default.
+Use can enable `rtsps` (TLS) by adding a `certificate = "/path/to/pem"` to the config. This pem should contain by the certificate and the key used for the server. If TLS is enabled all connections must use `rtsps`. You can also control client side TLS with the config option `tls_client_auth = "none|request|require"`, in this case the client should present a certificate signed by the server's CA, this is disabled by default.
 
 Use can enable basic authentication by adding the configuration option `username = "someone"` and `password = "somepass"`. Both username and password must be set for this to work.
 
--- a/README.md
+++ b/README.md
@@ -108,6 +108,10 @@ you also need to add the allowed users into each camera by adding the following 
 permitted_users = ["someone", "someoneelse"]
 ```
 
+Anywhere a username is accepted it can take any username or one of the following special values.
+- `anyone` means any user with a valid user/pass
+- `anonymous` means no user/pass required
+
 You can change the Neolink log level by setting the `RUST_LOG` environment variable (not in the configuration file) to one of `error`, `warn`, `info`, `debug`, or `trace`:
 
 ```
--- a/README.md
+++ b/README.md
@@ -112,6 +112,8 @@ Anywhere a username is accepted it can take any username or one of the following
 - `anyone` means any user with a valid user/pass
 - `anonymous` means no user/pass required
 
+The default `permitted_users` list is `[ "anonymous" ]` meaning no authentication required. If you want to use authenitcation then after adding your users through `[[users]]` please ensure you update the `permittted_users` for each camera.
+
 You can change the Neolink log level by setting the `RUST_LOG` environment variable (not in the configuration file) to one of `error`, `warn`, `info`, `debug`, or `trace`:
 
 ```
--- a/README.md
+++ b/README.md
@@ -93,9 +93,11 @@ You may also need to use `h264` format while using the SD stream.
 By default Neolink serves on all IP addresses on port 8554.
 You can modify this by changing the `bind` and the `bind_port` parameter.
 
-Use can enable `rtsps` (TLS) by adding a `certificate = "/path/to/pem"` to the config. This pem should contain by the certificate and the key used for the server. If TLS is enabled all connections must use `rtsps`. You can also control client side TLS with the config option `tls_client_auth = "none|request|require"`, in this case the client should present a certificate signed by the server's CA, this is disabled by default.
+You can enable `rtsps` (TLS) by adding a `certificate = "/path/to/pem"` to the top section of the config file. This PEM should contain by the certificate and the key used for the server. If TLS is enabled all connections must use `rtsps`. You can also control client side TLS with the config option `tls_client_auth = "none|request|require"`; in this case the client should present a certificate signed by the server's CA.
 
-Use can enable basic authentication by adding users to the configuration file as:
+TLS is disabled by default.
+
+You can password-protect the Neolink server by adding `[[users]]`` sections to the configuration file:
 ```
 [[users]]
 name: someone
--- a/src/config.rs
+++ b/src/config.rs
@@ -105,7 +105,7 @@ fn default_tls_client_auth() -> String {
 }
 
 fn default_permitted_users() -> Vec<String> {
-    vec!["anyone".to_string()]
+    vec!["anonymous".to_string()]
 }
 
 fn default_users() -> Vec<UserConfig> {
--- a/src/config.rs
+++ b/src/config.rs
@@ -30,8 +30,8 @@ pub struct Config {
 
     #[validate(regex(
         path = "RE_TLS_CLIENT_AUTH",
-        message = "Incorrect stream format",
-        code = "format"
+        message = "Incorrect tls auth",
+        code = "tls_client_auth"
     ))]
     #[serde(default = "default_tls_client_auth")]
     pub tls_client_auth: String,
--- a/src/config.rs
+++ b/src/config.rs
@@ -28,7 +28,11 @@ pub struct Config {
     #[serde(default = "default_certificate")]
     pub certificate: Option<String>,
 
-    #[validate(regex(path = "RE_TLS_CLIENT_AUTH", message = "Incorrect stream format", code = "format"))]
+    #[validate(regex(
+        path = "RE_TLS_CLIENT_AUTH",
+        message = "Incorrect stream format",
+        code = "format"
+    ))]
     #[serde(default = "default_tls_client_auth")]
     pub tls_client_auth: String,
 
@@ -112,8 +116,8 @@ fn default_users() -> Vec<UserConfig> {
     vec![]
 }
 
-fn validate_username(name: &str)  -> Result<(), ValidationError> {
-    let reserved_names = vec!("anyone", "anonymous");
+fn validate_username(name: &str) -> Result<(), ValidationError> {
+    let reserved_names = vec!["anyone", "anonymous"];
     if reserved_names.contains(&name) {
         return Err(ValidationError::new("This is a reserved username"));
     }
--- a/src/config.rs
+++ b/src/config.rs
@@ -31,6 +31,7 @@ pub struct Config {
     pub tls_client_auth: String,
 
     #[validate]
+    #[serde(default = "default_users")]
     pub users: Vec<UserConfig>,
 }
 
@@ -95,4 +96,8 @@ fn default_tls_client_auth() -> String {
 
 fn default_permitted_users() -> Vec<String> {
     vec!["anyone".to_string()]
+}
+
+fn default_users() -> Vec<UserConfig> {
+    vec![]
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -8,6 +8,7 @@ use regex::Regex;
 lazy_static! {
     static ref RE_STREAM_FORM: Regex = Regex::new(r"^([hH]26[45]|[ \t]*[!].*)$").unwrap();
     static ref RE_STREAM_SRC: Regex = Regex::new(r"^(mainStream|subStream|both)$").unwrap();
+    static ref RE_TLS_CLIENT_AUTH: Regex = Regex::new(r"^(none|request|require)$").unwrap();
 }
 
 #[derive(Debug, Deserialize, Validate, Clone)]
@@ -21,6 +22,13 @@ pub struct Config {
     #[validate(range(min = 0, max = 65535, message = "Invalid port", code = "bind_port"))]
     #[serde(default = "default_bind_port")]
     pub bind_port: u16,
+
+    #[serde(default = "default_certificate")]
+    pub certificate: String,
+
+    #[validate(regex(path = "RE_TLS_CLIENT_AUTH", message = "Incorrect stream format", code = "format"))]
+    #[serde(default = "default_tls_client_auth")]
+    pub tls_client_auth: String,
 }
 
 #[derive(Debug, Deserialize, Validate, Clone)]
@@ -58,4 +66,12 @@ fn default_format() -> String {
 
 fn default_stream() -> String {
     "both".to_string()
+}
+
+fn default_certificate() -> String {
+    "".to_string()
+}
+
+fn default_tls_client_auth() -> String {
+    "none".to_string()
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,15 +1,16 @@
 use serde::Deserialize;
 use std::net::SocketAddr;
 use std::time::Duration;
+use std::clone::Clone;
 use validator::Validate;
 use regex::Regex;
 
 lazy_static! {
     static ref RE_STREAM_FORM: Regex = Regex::new(r"^([hH]26[45]|[ \t]*[!].*)$").unwrap();
-    static ref RE_STREAM_SRC: Regex = Regex::new(r"^(mainStream|subStream)$").unwrap();
+    static ref RE_STREAM_SRC: Regex = Regex::new(r"^(mainStream|subStream|both)$").unwrap();
 }
 
-#[derive(Debug, Deserialize, Validate)]
+#[derive(Debug, Deserialize, Validate, Clone)]
 pub struct Config {
     pub cameras: Vec<CameraConfig>,
 
@@ -21,7 +22,7 @@ pub struct Config {
     pub bind_port: u16,
 }
 
-#[derive(Debug, Deserialize, Validate)]
+#[derive(Debug, Deserialize, Validate, Clone)]
 pub struct CameraConfig {
     pub name: String,
 
@@ -55,5 +56,5 @@ fn default_format() -> String {
 }
 
 fn default_stream() -> String {
-    "mainStream".to_string()
+    "both".to_string()
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -12,6 +12,7 @@ lazy_static! {
 
 #[derive(Debug, Deserialize, Validate, Clone)]
 pub struct Config {
+    #[validate]
     pub cameras: Vec<CameraConfig>,
 
     #[serde(rename = "bind", default = "default_bind_addr")]
--- a/src/config.rs
+++ b/src/config.rs
@@ -26,6 +26,12 @@ pub struct Config {
     #[serde(default = "default_certificate")]
     pub certificate: String,
 
+    #[serde(default = "default_username")]
+    pub username: String,
+
+    #[serde(default = "default_password")]
+    pub password: String,
+
     #[validate(regex(path = "RE_TLS_CLIENT_AUTH", message = "Incorrect stream format", code = "format"))]
     #[serde(default = "default_tls_client_auth")]
     pub tls_client_auth: String,
@@ -74,4 +80,12 @@ fn default_certificate() -> String {
 
 fn default_tls_client_auth() -> String {
     "none".to_string()
+}
+
+fn default_username() -> String {
+    "".to_string()
+}
+
+fn default_password() -> String {
+    "".to_string()
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -24,7 +24,7 @@ pub struct Config {
     pub bind_port: u16,
 
     #[serde(default = "default_certificate")]
-    pub certificate: String,
+    pub certificate: Option<String>,
 
     #[serde(default = "default_username")]
     pub username: Option<String>,
@@ -74,8 +74,8 @@ fn default_stream() -> String {
     "both".to_string()
 }
 
-fn default_certificate() -> String {
-    "".to_string()
+fn default_certificate() -> Option<String> {
+    None
 }
 
 fn default_tls_client_auth() -> String {
--- a/src/config.rs
+++ b/src/config.rs
@@ -27,10 +27,10 @@ pub struct Config {
     pub certificate: String,
 
     #[serde(default = "default_username")]
-    pub username: String,
+    pub username: Option<String>,
 
     #[serde(default = "default_password")]
-    pub password: String,
+    pub password: Option<String>,
 
     #[validate(regex(path = "RE_TLS_CLIENT_AUTH", message = "Incorrect stream format", code = "format"))]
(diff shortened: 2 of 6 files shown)


QuantumEntangledAndy:
Build seems to be failing in chocolaty timing out when downloading the gstramer msi.

thirtythreeforty:
Nice, you're on a roll here. Thanks again. Let me review in a little.

Yeah don't feel bad about the CI, GitHub Actions seems to have problems quite routinely. We'll rerun them.

QuantumEntangledAndy:
Yes I am having fun with rust and my little E1 camera.

src/gst.rs
```diff
@@ -84,21 +98,47 @@ impl RtspServer {
...
-        });
+    pub fn set_credentials(&self, user :&str, pass :&str) -> Result<()> {
+        let auth = match self.server.get_auth() {
+            Some(x) => x,
+            None =>  RTSPAuth::new(),
+        };
+
+        if ! user.is_empty() && ! pass.is_empty() {
+            info!("Setting credentials for user {}", user);
+            debug!("Password is {}", pass);
```
thirtythreeforty, line 109: Never print passwords into a log. Change "setting credentials" to `debug` log level.

src/gst.rs
```diff
@@ -84,21 +98,47 @@ impl RtspServer {
...
+        Ok(())
+    }
+
+    pub fn set_tls(&self, cert_file: &str, client_auth: TlsAuthenticationMode) -> Result<()> {
+        if ! cert_file.is_empty() {
+            info!("Setting up TLS using {}", cert_file);
+            let auth = match self.server.get_auth() {
+                Some(x) => x,
+                None =>  RTSPAuth::new(),
+            };
```
thirtythreeforty, line 133: The idiomatic way to unwrap or use a default is `Option::unwrap_or(default_value)`.

Then, because constructing a default that doesn't get used can be expensive, there's also the version that accepts a closure, `unwrap_or_else`.  So this should be:

```rust
let auth = self.server.get_auth().unwrap_or_else(RTSPAuth::new);
```

src/gst.rs
```diff
@@ -84,21 +98,47 @@ impl RtspServer {
...
+        }
 
-        self.server.set_auth(auth.as_ref());
+        self.server.set_auth(Some(&auth));
+        Ok(())
+    }
+
+    pub fn set_tls(&self, cert_file: &str, client_auth: TlsAuthenticationMode) -> Result<()> {
+        if ! cert_file.is_empty() {
+            info!("Setting up TLS using {}", cert_file);
```
thirtythreeforty, line 129: Again, `debug` level here

src/gst.rs
```diff
@@ -84,21 +98,47 @@ impl RtspServer {
...
+            let auth = match self.server.get_auth() {
+                Some(x) => x,
+                None =>  RTSPAuth::new(),
+            };
+
+            let cert = TlsCertificate::new_from_pem(&fs::read_to_string(cert_file).expect("TLS file not found")).expect("Not a valid TLS certificate");
```
thirtythreeforty, line 135: Looks like there's `new_from_file` which will avoid having to slurp the whole thing into a string.

src/main.rs
```diff
@@ -51,6 +52,16 @@ fn main() -> Result<(), Error> {
 
     let rtsp = &RtspServer::new();
 
+    let tls_client_auth = match &config.tls_client_auth as &str {
+        "request" => TlsAuthenticationMode::Requested,
+        "require" => TlsAuthenticationMode::Required,
+        "none"|_ => TlsAuthenticationMode::None,
+    };
```
thirtythreeforty, line 59: I'm not a fan of defaulting `_` to None. Say there was a typo `reqiure`. The user doesn't intend for it to be none.

Now, you do have a regex protecting this. So make the default match branch go

```rust
_ => unreachable!()
```

which is a macro that will panic if it's ever hit but documents that you think it's unreachable.

thirtythreeforty: Overall I think a cleaner way to do one-of config values like this would be to have an enum and let the config parser figure out which variant it is, but I don't have much experience with that so this approach is fine.

src/gst.rs
```diff
@@ -84,21 +98,47 @@ impl RtspServer {
         Ok(maybe_app_src)
     }
 
-    pub fn set_credentials(&mut self, user_pass: Option<(&str, &str)>) -> Result<()> {
-        let auth = user_pass.map(|(user, pass)| {
```
thirtythreeforty, line 88: I like the original function signature with user/pass being passed in together. That way you know the username/password both exist or neither exist when you arrive in the function, and validating the combo stays in main or config parsing or wherever.

Also, if you have an unspecified user or pass in the config, you'd want it to be `None`, not `""`. So change the config value to `Option<String>`, which should let you get rid of the associated default method.

In other words, keep the boundary between "stringly typed" config and strongly-typed Rust at the first opportunity.

src/config.rs
```diff
@@ -21,6 +22,19 @@ pub struct Config {
     #[validate(range(min = 0, max = 65535, message = "Invalid port", code = "bind_port"))]
     #[serde(default = "default_bind_port")]
     pub bind_port: u16,
+
+    #[serde(default = "default_certificate")]
+    pub certificate: String,
```
thirtythreeforty, line 27: I *think* you can use a `Path` here or similar and it will auto validate that it exists. But I could be thinking of a different library so ignore if so.

src/gst.rs
QuantumEntangledAndy: I did like your original too but I had trouble getting it to compile.

How would I get
```
let auth = user_pass.map(|(user, pass)| {
```
to work with the case of None
I still need the `auth` variable to set the default of allowing access.

If I don't let auth and instead do
```
user_pass.map(|(user, pass)| {
// stuff with pass
}
```
I still need to handle the none to setup default persmission

This is looking like it just needs a match but perhaps you have a better idea?

thirtythreeforty: You'd still get-or-create `auth` from the server object as you do now. After that, you'd optionally set the user/pass up if they exist:
```rust
if let Some((user, pass)) = user_pass {
    // Now user and pass are in scope, modify auth
} else {
    // They don't exist (user_pass == None) so do the default
}

File: src/config.rs (hunk 7 above)
Comment (QuantumEntangledAndy) [reply src/config.rs:27]: So tried using path, but first rust said it didn't have fixed size so I needed to use &, then it said I something `expected named lifetime parameter` and I think my rust needs a level up first haha

File: src/gst.rs (hunk 1 above)
Comment (QuantumEntangledAndy) [reply src/gst.rs]: Sure sounds like a good plan

File: src/gst.rs (hunk 2 above)
Comment (QuantumEntangledAndy) [reply src/gst.rs]: I learn a new thing, thankyou

File: src/gst.rs (hunk 3 above)
Comment (QuantumEntangledAndy) [reply src/gst.rs]: Done

File: src/gst.rs (hunk 4 above)
Comment (QuantumEntangledAndy) [reply src/gst.rs]: Ok, but this means that the user won't get a different message if its a missing file or if incorrect format.

File: src/main.rs (hunk 5 above)
Comment (QuantumEntangledAndy) [reply src/main.rs:60]: Yes I was thinking it didn't matter because of the regex but this is better

File: src/gst.rs (hunk 6 above)
Comment (QuantumEntangledAndy) [reply src/gst.rs:88]: Ok, got it working but not quite the same format.. please comment

File: src/config.rs (hunk 7 above)
Comment (QuantumEntangledAndy) [reply src/config.rs]: I've left it as string but I am using an Option for it in the config

**Comment by QuantumEntangledAndy** [comment]
So that should be everything, I still need to test with my E1

**Comment by QuantumEntangledAndy** [comment]
Also I have a basic implementation of multiple users and permissions per stream set up. Some of these changes will get reworked by that branch when it is finished.

File: src/gst.rs (hunk 4 above)
Comment (thirtythreeforty) [reply src/gst.rs]: Oh, that's a good point, especially relevant with TLS certs. The way you have it will be fine; can you add a comment explaining this?

File: src/config.rs (hunk 7 above)
Comment (thirtythreeforty) [reply src/config.rs]: Oh, yeah Path is a trait I think (kinda like `interface` in Java), which is not sized. You would want `OsPath` or something, I haven't used that area of the stdlib much. String is definitely fine.

**Comment by thirtythreeforty** [comment]
Cool. I guess the config file format will change - would you like to hold off merging this so it doesn't have to support the old format? We can also just break compatibility but it'd be nice to have old configs Just Work.

File: src/gst.rs (hunk 6 above)
Comment (QuantumEntangledAndy) [reply src/gst.rs:88]: Oh that's cleaner I will change it to the method you suggested

**Comment by QuantumEntangledAndy** [comment]
Um the changes to the config file format were going to
- Add Vec<UserConfig>
- Add UserConfig structure with user and pass
- Add permitted_users to camara config

So yes we will loose the `username` and `password` config introduced in this config.

I could leave them in as a sort of master user that can access all streams

File: src/gst.rs (hunk 6 above)
Comment (thirtythreeforty) [reply src/gst.rs:88]: Yeah you get this gist, `if let` is sugar over `match`.

File: src/gst.rs [hunk 8]
```diff
@@ -84,21 +97,42 @@ impl RtspServer {
...
+    }
+
+    pub fn set_tls(&self, cert_file: &str, client_auth: TlsAuthenticationMode) -> Result<()> {
+        debug!("Setting up TLS using {}", cert_file);
+        let auth = self.server.get_auth().unwrap_or_else(|| RTSPAuth::new());
 
-        self.server.set_auth(auth.as_ref());
+        let cert = TlsCertificate::new_from_file(&cert_file).expect("Not a valid TLS certificate or not found.");
+        auth.set_tls_certificate(Some(&cert));
+        auth.set_tls_authentication_mode(client_auth);
```
thirtythreeforty, line 129: Does this line prohibit TLS _and_ basic auth? I would think that's a valid combination.

QuantumEntangledAndy: By default when you turn on TLS you get auth turned on that stops you access anything unauthorized.

- We first set up TLS which stop everything
- When setting up the factory we add permissions so that anyone with the watcher token to play
- We then setup users
   - if there is a username we permit them only by giving only them the `watcher` token
   - If there is no username we permit anyone we give the default user the `watcher` token so that all can watch

I have set it up so that  you are able to choose TLS or Basic or TLS+Basic.

I may have missunderstood your question... Where you actually referring to `auth.set_tls_authentication_mode(client_auth);`

thirtythreeforty:
> Um the changes to the config file format were going to
> 
>     * Add `Vec<UserConfig>`
> 
>     * Add `UserConfig` structure with user and pass
> 
>     * Add permitted_users to `CamaraConfig` which is a `Vec<String>` of user names
> 
> 
> So yes we will loose the `username` and `password` config introduced in this config.
> 
> I could leave them in as a sort of master user that can access all streams

Great approach for user list.  In toml this would look like:
```
[[users]]
username = thirtythreeforty
password = hunter2
```

I'd say if you have that working or nearly working, let's just go with that version so the standalone user/pass doesn't get left over.

QuantumEntangledAndy:
Ok I will finish up that one pulling in these changes and then merge into this PR branch

src/gst.rs
QuantumEntangledAndy:
```
auth.set_tls_authentication_mode(client_auth);
``
As I understand this... possible wrong:
- `client_auth = required`
    - Means I want the client to have a PEM certificate too. That cert must be signed by me. No cert from me, no access.
    - This is on top of basic auth, and adds extra security. I like to think of it as the equlivalent of an `sshkey` being required by the server.
- `client_auth = requested`
    - Means I accept PEMs signed by me or someone else, you don't need one but it would be nice.
- `client_auth = none`
    - I don't need to PEM at all and won't bother checking for one.

I am not sure of all the details about `requested` vs `none`, the docs are about sparse.

File: src/gst.rs (hunk 4 above)
Comment (QuantumEntangledAndy) [reply src/gst.rs]: Ok changed back being extra verbose about the reason

**Comment by QuantumEntangledAndy** [comment]
So this is the multiple users branch pulled in, only issue is my E1 is offline... So I cannot test it until I get home. Is there a way to ignore `Error streaming from camera name, will retry in...`  and just pull data from say the gstreamer testsrc

**Comment by QuantumEntangledAndy** [comment]
Ah lack of connection may have something to do with your changes to the mount points. I specify `stream = "subStream"` and I was expecting it to mount on `name` not `name/subStream` its a bit different to what I implemented.

p.s. Does substream work on your camera, I thought it needed to be H264? There's no way to specify separate formats when using "both". It used to force it to always be H264 on the `subStream` but this is gone now. Will you be adding auto detection :)?

File: src/gst.rs (hunk 8 above)
Comment (thirtythreeforty) [reply src/gst.rs:132]: I think you're correct about the `client_auth` semantics. I was referring to the line below, apologies. You run
```rust
auth.set_supported_methods(RTSPAuthMethod::None);
```
which, I would think that auth method is independent of whether the connection is encrypted. For example you could have an encrypted unauthenticated connection (like most HTTPS websites) or an unencrypted authenticated connection (which would be ill-advised because the passwords could be sniffed).

thirtythreeforty: In other words, let the auth-setup set the auth up, and let the encryption-setup set the encryption up.

src/main.rs
```diff
@@ -127,6 +133,56 @@ struct CameraErr {
...
+
+fn set_up_users(users: &Vec<UserConfig>, rtsp: &RtspServer) {
+    // Setting up users
+    let mut credentials = vec![];
+    for user in users {
+        let name = &user.name;
+        let pass = &user.pass;
+        let user_pass = match (name, pass)  {
+            (Some(name), Some(pass)) => Some((&name as &str, &pass as &str)),
+            _ => None,
```
thirtythreeforty, line 156: Might be worth a `warn!()` here if you get `(Some(), None)` or `(None, Some())` to let the user know they dun goofed.

src/gst.rs
thirtythreeforty: > Ah lack of connection may have something to do with your changes to the mount points. I specify `stream = "subStream"` and I was expecting it to mount on `name` not `name/subStream` its a bit different to what I implemented.

Didn't intentionally change the semantics, but I also like having each stream at a stable path. Sorry for the confusion.
 
> p.s. Does substream work on your camera, I thought it needed to be H264? There's no way to specify separate formats when using "both". It used to force it to always be H264 on the `subStream` but this is gone now. Will you be adding auto detection :)?

Yes it did and yes I will, soon (tm). It doesn't surprise me that there's a single codec per camera; usually the SoCs they use have one powerful encoder and that's all.

thirtythreeforty:
> Ah lack of connection may have something to do with your changes to the mount points. I specify stream = "subStream" and I was expecting it to mount on name not name/subStream its a bit different to what I implemented.

Didn't intentionally change the semantics, but I also like having each stream at a stable path. Sorry for the confusion.

> p.s. Does substream work on your camera, I thought it needed to be H264? There's no way to specify separate formats when using "both". It used to force it to always be H264 on the subStream but this is gone now. Will you be adding auto detection :)?

Yes it did and yes I will, soon (tm). It doesn't surprise me that there's a single codec per camera; usually the SoCs they use have one powerful encoder and that's all.

src/main.rs
thirtythreeforty: Rule of thumb: every warning, every sanity check, will save you from dealing with at least one trouble ticket.

QuantumEntangledAndy:
That's seems reasonable but maybe need to explain the paths in docs a bit more.

With regards to multiple users I have something odd happening.

As far as I can tell I've set it up in the correct way.

- Make a default token for anonymous that has the noauth role
- Make a token for each user that has its own role name
- On each factory add permissions for the each allowed role name

The interesting things happen at runtime...

- If I have the default token setup all users are using that role. The role I give it is ignored. This means Auth fails unless anonymous is allowed. In which case why bother. 
- If I comment out the bit about setting up the default token then each user has their own tokens.

I believe one of the following is happening

- When a new token is added it combines it with the default but drops any shared fields from the new one. A check of the gstramer source seems to suggest this is not the case.

- During authentication the client first connects anonymously and gets their tokens. After authentication it gains the user's token but drops shared fields again. This may be client dependant, I've been testing with ffmpeg which is usually very standard.
    - Might be worth testing in VLC

- I am not doing it correctly. Since my code is quite similar to the example in the gstramer source I am hoping it's not this.

- Theres an upstream bug where it is not dropping old fields after authentication. I couldn't see such a thing in the gstramer source but I have yet to check the rust port.

So all in all I need more time to resolve this. I shall be not available for a few days with life things so maybe have to wait to track this down.

src/gst.rs
QuantumEntangledAndy: Yes that was the general idea setting up basic should be done in the Auth set up. Originally I was hoping that setting it to none would restore the default functionality of unauthenticated can do anything but that was not the case. Having tls and no auth without  setting up default token and presmissions in factory, means we are encrypted but no-one ace access it haha. You have to always run the user set up and factory permissions with tls.

src/main.rs
QuantumEntangledAndy: That means a warn and a return None. Would that need an enclosure I wonder or something else. More rust foo to learn in the docs :).

thirtythreeforty: Yeah, you would do:
```rust
(Some(_), None) | (None, Some(_)) => {
    warn!("Username and password must be supplied together - ignoring [[users]] entry");
    None
}
```

because the block evaluates to the last statement in it.

thirtythreeforty:
So when I modified your previous PR I totally hosed mainstream/substream; it was trying to connect to the bogus `both` stream. This is fixed in 94be9e603c0a6dfd09a5a03bf13934ad18a3c72c.  You were right about substream always being H.264. Make sure you merge master into your working branch to pick up the fix. Hope this didn't cause too much grief.

QuantumEntangledAndy:
It's weird took have two encoders but I thought it must be because you're previous issue described it as not working and was maybe in h264. 

It didn't cause any grief but I did learn how to make a dummy camera that dosent try connecting which is helping my debugging without always compiling on my rasberry for every test. 

Anyways with regards to multiple users I think I can do this:
- If no users are specified allow all anonymously
- If users specified no anonymous binding at all (I originally tried to allow this but gstreamer was fighting me)

This is more involved code wise as it required special handling of the no users case.

Still can't work on this but hopefully soon

src/main.rs
QuantumEntangledAndy: I have added this in but it should also be caught by the validation

QuantumEntangledAndy:
Ok so I am mostly back, I have gotten it to work as I originally intended.

- Users are supplied with
```
[[users]]
name = "me"
pass = "mepass"
```
in the config

- Each camera has an option in the config of
```
permitted_users = [ "me", "someone" ]
```
that can choose who can access.

- If `permitted_users` is not included it default to [ "anyone" ] including "anonymous" binding.

src/gst.rs
QuantumEntangledAndy: I think I should remove it from the TLS part as it's not really helping.

QuantumEntangledAndy: Ok so I removed the setting of basic to None during the TLS. Everything works the same even without it.

src/main.rs
```diff
@@ -131,6 +137,69 @@ struct CameraErr {
...
+    // If in the camera config there is the user "anyone"
+    // Then we add all users to the cameras config. including unauth
+    let mut new_permitted_users = vec![];
+    if current_permitted_users.contains(&"anyone".to_string()) {
+        for credentials in users {
+            if let Some(user) = &credentials.name {
+                new_permitted_users.push(user.to_string());
+            }
+        }
+        new_permitted_users.push("anonymous".to_string());
```
thirtythreeforty, line 192: Ooh so in my head "anonymous" is distinct from "anyone". Let's nail the semantics down. To me:

- `anyone` means "any properly authenticated user"
- `anonymous` means "any connecting user"

Do you agree?

README.md
```diff
@@ -93,6 +93,19 @@ You may also need to use `h264` format while using the SD stream.
 By default Neolink serves on all IP addresses on port 8554.
 You can modify this by changing the `bind` and the `bind_port` parameter.
 
+Use can enable `rtsps` (TLS) by adding a `certificate = "/path/to/pem"` to the config. This pem should contain by the certificate and the key used for the server. If TLS is enabled all connections must use `rtsps`. You can also control client side TLS with the config option `tls_client_auth = "none|request|require"`, in this case the client should present a certificate signed by the server's CA, this is disabled by default.
+
+Use can enable basic authentication by adding users to the configuration file as:
```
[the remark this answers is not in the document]
thirtythreeforty: We are rapidly trending toward needing a Configuration File Reference Manual

README.md
```diff
@@ -93,6 +93,19 @@ You may also need to use `h264` format while using the SD stream.
...
+```
+[[users]]
+name: someone
+pass: somepass
+```
+you also need to add the allowed users into each camera by adding the following to `[[cameras]]`.
```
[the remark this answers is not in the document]
thirtythreeforty: Never mind, I see `anyone` is designed for this.

thirtythreeforty:
By the way - at any point feel free to indicate you'd like for me to run with it. Don't let me wear you out with feedback; I really appreciate your contributions so far.

QuantumEntangledAndy:
Sure if it gets to that point I will let you know but I don't really mind. I am having fun with the coding. Just can't always work on it

src/main.rs
```diff
@@ -131,6 +137,69 @@ struct CameraErr {
...
+}
+
+fn set_up_users(users: &Vec<UserConfig>, rtsp: &RtspServer) {
+    // Setting up users
+    let mut credentials = vec![];
+    let reserved_names = vec!(String::from("anyone"), String::from("anonymous"));
+    for user in users {
+        if let Some(name) = &user.name {
+            if reserved_names.contains(name) {
+                warn!("{} is a reserved username, skipping this user", name);
```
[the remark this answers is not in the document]
QuantumEntangledAndy: Failing this way with a useful error message might mean using a custom validator. I will have to look into how that is done

src/main.rs
```diff
@@ -131,6 +137,69 @@ struct CameraErr {
...
+                new_permitted_users.push(user.to_string());
+            }
+        }
+        new_permitted_users.push("anonymous".to_string());
+    } else {
+        new_permitted_users.append(&mut current_permitted_users.clone());
+    }
+    // We also drop duplicated user names
+    new_permitted_users.sort();
+    new_permitted_users.dedup();
```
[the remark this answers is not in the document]
QuantumEntangledAndy: I was thinking of `set` when I wrote this but I found this solution in my googling first so just rolled with it :)

src/main.rs
QuantumEntangledAndy: That sounds like a reasonable method. I think I should change the default to `anonymous` though so that it won't effect current configs.

src/main.rs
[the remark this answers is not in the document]
QuantumEntangledAndy: Ok I've gotten a custom validator here, so it now works through the validation

src/main.rs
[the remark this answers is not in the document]
QuantumEntangledAndy: I have changed this to a hashset but should I pass a hashset in to the other places as an argument

README.md
```diff
@@ -93,6 +93,19 @@ You may also need to use `h264` format while using the SD stream.
 By default Neolink serves on all IP addresses on port 8554.
 You can modify this by changing the `bind` and the `bind_port` parameter.
 
+Use can enable `rtsps` (TLS) by adding a `certificate = "/path/to/pem"` to the config. This pem should contain by the certificate and the key used for the server. If TLS is enabled all connections must use `rtsps`. You can also control client side TLS with the config option `tls_client_auth = "none|request|require"`, in this case the client should present a certificate signed by the server's CA, this is disabled by default.
```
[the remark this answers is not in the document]
QuantumEntangledAndy: Ok done

README.md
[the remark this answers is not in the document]
QuantumEntangledAndy: Yes perhaps we should add comments to the examples config toml

src/gst.rs
```diff
@@ -32,7 +35,7 @@ impl RtspServer {
         }
     }
 
-    pub fn add_stream(&self, paths: &[&str], stream_format: &StreamFormat) -> Result<MaybeAppSrc> {
+    pub fn add_stream(&self, paths: &[&str], stream_format: &StreamFormat, permitted_users: &Vec<String>) -> Result<MaybeAppSrc> {
```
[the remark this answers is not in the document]
QuantumEntangledAndy: I see, should we change this to a HashSet though

src/main.rs
QuantumEntangledAndy: Or should we have `anonymous` if no users in the config or `anyone` when users are in the config?

QuantumEntangledAndy: The problem with that way though is how to set up the condition in the toml default function...

QuantumEntangledAndy:
p.s. why do I keep getting emails about `Run failed: Publish Docker Image...` surely we shouldn't try to publish a docker image off of the pull request branches

src/main.rs
QuantumEntangledAndy: So I set it up like this, permitted_users is now an option which defaults to None.

In the `get_permitted_users` function we check for None or Some
- if Some then use what is in the toml
- if None check
  - if length of users>0  then use `[ "anyone" ]`
  - if length of users == 0 use `[ "anonymous" ]`

QuantumEntangledAndy: So I set it up like this. `permitted_users` is now an Option which defaults to None

In the function `get_permitted_users` we match `permitted_users` to
- Some, use what was in the toll
- None, we check the length of users I.e. number of `[[users]]`
  - If length > 0 `permited_users` defaults to `[ "anyone" ]`. Meaning any of the `[[users]]` can connect but not anonymous. This means you can activate authentication with just a `[[users]]` block.
  - If length == 0 `permited_users` defaults to `[ "anoymous" ]`. Meaning anyone even without authentication. This reflects the current default so no changes in config required.

QuantumEntangledAndy:
> Thanks for working on this! I'm glad you're tackling it, this is more complex than I had realized.

It is more difficult than I thought too, mostly due to gstreamers lack of compete documentation or examples. I was hoping this would be a nice easy one to tackle before I tried something more complex like polling the motion detection from the camera haha.

thirtythreeforty:
> p.s. why do I keep getting emails about `Run failed: Publish Docker Image...` surely we shouldn't try to publish a docker image off of the pull request branches

When you push to your fork, you get errors because you don't have Docker credentials set up. #38. Feel free to ignore till a better solution is in place. I only touch the Docker stuff when I absolutely have to, ha.

src/main.rs
thirtythreeforty: Yep, I think that's the behavior I'd expect. Sounds good.

README.md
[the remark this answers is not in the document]
thirtythreeforty: That is a very good idea. Thanks.

src/gst.rs
```diff
@@ -84,21 +104,71 @@ impl RtspServer {
...
+        let auth = self.server.get_auth().unwrap_or_else(|| RTSPAuth::new());
+        auth.set_supported_methods(RTSPAuthMethod::Basic);
+
+        let mut un_authtoken = RTSPToken::new(&[(*RTSP_TOKEN_MEDIA_FACTORY_ROLE, &"anonymous")]);
+        auth.set_default_token(Some(&mut un_authtoken));
+
+        for credential in credentials {
+            if let Some((user, pass)) = credential {
+                debug!("Setting credentials for user {}", user);
+                debug!("Password is {}", pass);
```
thirtythreeforty, line 149: Please do not print passwords

src/config.rs
```diff
@@ -23,6 +24,21 @@ pub struct Config {
     #[validate(range(min = 0, max = 65535, message = "Invalid port", code = "bind_port"))]
     #[serde(default = "default_bind_port")]
     pub bind_port: u16,
+
+    #[serde(default = "default_certificate")]
+    pub certificate: Option<String>,
+
+    #[validate(regex(
+        path = "RE_TLS_CLIENT_AUTH",
+        message = "Incorrect stream format",
```
thirtythreeforty, line 33: Copypasta here?

src/main.rs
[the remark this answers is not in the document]
thirtythreeforty: Yes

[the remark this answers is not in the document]
QuantumEntangledAndy: Ok done, although I had to come up with something different for joining the hashmap into a string during the debug pint

src/gst.rs
QuantumEntangledAndy: Ok done

src/config.rs
QuantumEntangledAndy: Umm yes.. fixed

thirtythreeforty:
Merged, thank you very much again.  I refactored the setup function a little so it passes `&[&str]` around where possible.

QuantumEntangledAndy:
I'm glad it is working alright :). I'll check out your edits and see if I can learn a bit more rust.

PR #26, opened 2020-07-03
Implement Docker image based on Alpine

### LINKED ISSUE #16: Have created a docker image for app
Pretty Basic based off the ubuntu base image (rather than source - optimised version will be coming soon built from source and latest code)

docker run  --rm -p 9554:8554  -v <location of your>/neolink.conf:/etc/neolink.conf mr2jzgte/neolink

exposed on port 9554, modify as required.

if needed, troubleshoot via 

docker run  --rm -it -p 9554:8554 --entrypoint bash -v <location of your>/neolink.conf:/etc/neolink.conf mr2jzgte/neolink

Have a builtin watchdog, that if the process isn't running, it will restart it. (may need tweaking)
My first docker image build - so it's a learning curve

thirtythreeforty:
Still in the middle of testing. Thanks @thadius83 for the great starting point (I kinda don't like dealing with Docker so it was nice). I think your goal was to get it smaller and build automatically; I've done my best here.

Closes #16

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+target/
+Dockerfile
+.dockerignore
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,0 +1,19 @@
+# A single-shot build-n-pack on Alpine Linux
+name: Publish Docker image
+on: [push]
+
+jobs:
+  push_to_registry:
+    name: Build Docker image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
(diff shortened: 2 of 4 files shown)


thirtythreeforty:
Closes #26

thirtythreeforty:
**_News Flash_**
In a stunning development that software engineers are calling "super dumb" and "amazingly useless," it has been [reported that users cannot pull from public GitHub Docker registries without logging in](https://github.community/t/docker-pull-from-public-github-package-registry-fail-with-no-basic-auth-credentials-error/16358).

PR #40, opened 2020-07-15
Add a build action for armhf

### LINKED ISSUE #27: Provide ARMv6 and ARMv7 Linux builds
Many people want to use Neolink on a Raspberry Pi. This is perfectly feasible. However, it would take a long time to compile. Instead, provide a cross-compiled prebuilt for ARM.

QuantumEntangledAndy:
So this PR adds a build action for armhf

It is mostly build off of [cargo cross](https://github.com/rust-embedded/cross) which is a very nice way of cross compiling rust projects

The default docker image for armhf from `cross` was however not working for us (because of the gstreamer dependencies) so I had to build a new docker based off of the one from cargo cross.

Currently the docker image is built *EVERY* time the action is run. Which is not ideal, really it should be published and then pulled when needed since that that part never changes. But I thought I would put the PR here so you can see what you think before I try anything else more complex.

fix #27 (well armv7 anyways)

### DIFF
--- a/.github/workflows/armhf.yml
+++ b/.github/workflows/armhf.yml
@@ -1,0 +1,51 @@
+name: Build ARMHF
+
+on: [push, pull_request]
+
+jobs:
+  armhfbuild:
+    name: build-armhf
+    runs-on: ubuntu-latest
+    container: node:current-buster-slim
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install basic tools
+        run: |
+          apt-get update
+          apt-get install --assume-yes --no-install-recommends sudo curl ca-certificates build-essential
+      - name: Add armhf as a foreign architecture
+        run: |
+          sudo dpkg --add-architecture armhf
+      - name: Install cross compile tools
+        run: |
+          sudo apt-get update
+          sudo apt-get install --assume-yes --no-install-recommends gcc-arm-linux-gnueabihf
+      - name: Install armhf gstreamer
+        run: |
+          sudo apt-get update
+          sudo apt-get install --assume-yes --no-install-recommends \
+          libgstrtspserver-1.0-dev:armhf \
+          libgstreamer1.0-dev:armhf \
+          libgtk2.0-dev:armhf
+      - name: Install rustup armv7
+        uses: actions-rs/toolchain@v1
+        with:
+            toolchain: stable
+            override: true
+            target: armv7-unknown-linux-gnueabihf
+      - name: Build armfh
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --release --all-features --target=armv7-unknown-linux-gnueabihf
+        env:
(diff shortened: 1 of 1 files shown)


QuantumEntangledAndy:
So this builds on GitHub and then works on my buster raspberry pi just find.

It is for armhf (armv7).

To get an armv6 we would need to change it to use armel I think. I think buster supports this so the gstramer packages are in there. But we do have the issue that cross uses qemu which dosent support armv6. I think qemu is used for running the code tests not compling though so it may work. This would mean another docker image though and I don't have an armv6 to test it on.

QuantumEntangledAndy:
So I have now made a  version that dosen't use cross or a custom docker image, it runs in github actions. If you would prefer that please let me know and I will merge that branch into this.

I was thinking to get armv6 aswell but ran into more issues that might be fixable using https://github.com/Pro/raspi-toolchain. The issue is that modern compilers don't build their toolkits for armv6.

thirtythreeforty:
Thank you for this.

I peeked at your Github Actions branch and I really like the approach there. Can you merge it into this branch?

And yeah I'd say drop armv6 for now, most people will be running on newer stuff. The only recent Pi that's ARMv6 is the Zero and since it lacks an Ethernet port I think most people won't be bothering with that target.

QuantumEntangledAndy:
Sure I like that method too, seems much cleaner.

I might make another PR soon with the macos build

QuantumEntangledAndy:
Ok so I swapped it to using the other method (and squashed it to one commit)

thirtythreeforty:
@jlg89 would you want to download [the build artifact](https://github.com/thirtythreeforty/neolink/suites/978499887/artifacts/12490346) from this PR and try it on your Pi?

thirtythreeforty:
@QuantumEntangledAndy I guess this new branch runs on your Buster Pi as well? I am just looking for a :+1:  from someone before I hit merge.

QuantumEntangledAndy:
Well it works on my buster for sure, I have another rpi with stretch somewhere too if you want me to test there too

thirtythreeforty:
I think it is unlikely to work on stretch since the code is linked against newer libraries (and the old gstreamer wouldn't be forward compatible). But buster is stable and what RPi is shipping, so I'm fine with limiting the amount of effort expended here. Thanks for testing.

QuantumEntangledAndy:
If thats the case I might make a PR so we can have artifacts for stretch and buster

thirtythreeforty:
I certainly won't say no if you want to!

jlg89:
Sorry I went dark, had to sleep. It does run on my RPi. Won't connect to the camera (Argus Eco), but that's probably a different issue.

thirtythreeforty:
No worries, glad it works for you. Yeah, we have no support for Argus right now.

QuantumEntangledAndy:
Just wanted to add that I tried to make it work on stretch but alas we need a newer gst version than is provided:

```
Requested 'gstreamer-1.0 >= 1.14' but version of GStreamer is 1.10.4
```

So buster required I for the correct gst version

PR #44, opened 2020-07-21
addon/plugin/costume component For Home Assistant  - Neolink 

nunokaneco:
Hello All

It's possible to create a addon/plugin/costume component based in neolink to running in HA -Home Assistant 

I think this can be useful once we can integrate the reolink cameras that is not supporting the RTSP or ONVIF protocols. 

Converting in a supported devices to be available :) 

Think about tis please 

Best regards


thirtythreeforty:
Thanks for the request. Migrating to an issue, not a pull request - see #45.

Issue #46, opened 2020-07-23
Segmentation fault 


derkenblosh:
getting this up and running for the first time. using the hosted docker image.

I'm able to test in BI, and connect to the subStream without issue (i do get random Segmentation Faults, but they do not seems to effect my stream). But when i attempt to access the mainStream, i get repeated Segmentation Faults every few seconds. 

#https://github.com/thirtythreeforty/neolink                  
bind = "0.0.0.0"                                              
                                                              
[[cameras]]                                                   
name = "HD1"                                                  
username = "admin"                                            
password = ""                                         
address = "192.168.11.207:9000"                               
stream = "both"                                               
format = "H264"                                               
timeout = { secs = 5, nanos = 0 }                             
~                                                             
~                                                             
~                                                             
~                                                             
~                                                             
~                                                             
~                                                             
~                                                             
~                                                             
~                                                             
~                                                             
~                                                             
~                                                             
neolink.toml                                1,4            All

[2020-07-23T01:03:51Z INFO  neolink] HD1: Connected to camera, starting video stream mainStream,
Segmentation fault (core dumped),

derkenblosh:
... Going to setup on the host machine and report back if I receive the same errors. After some research, this looks like it might actually be a docker issue.

m1k1o:
@derkenblosh maybe releated to https://github.com/thirtythreeforty/neolink/issues/37? It looks like the same problem. Try to connect to stream explicitly using UDP. If it works, then it's that TCP issue.

thirtythreeforty:
I'd bet the root cause is Gstreamer handling the extra data from #49 and #50 incorrectly.  Granted it shouldn't crash, but I bet when that's implemented this problem will go away.

thirtythreeforty:
@m1k1o would you try the Docker image for the `magic_headers` branch?

```
docker pull thirtythreeforty/neolink:magic_headers
```

It still has implementation issues but I want to see if it outright crashes on you.

m1k1o:
@thirtythreeforty nope, still crashes. On top of that, UDP stream can't be watched. It's just gray, 100% loss of picture.

thirtythreeforty:
Well gross.

QuantumEntangledAndy:
@m1k1o If you have the time could you try my branch `quantumentangledandy/neolink:magic_headers`

```bash
docker pull quantumentangledandy/neolink:magic_headers

[...]

PR #50, opened 2020-07-28
Filter data sent to gsteamer using magic bytes

### LINKED ISSUE #49: Invalid NAL units reported by ffmpeg
**Describe the bug**
When using neolink with ffmpeg, ffmepg reports that it is recieving invalid NAL (h264) units.

**To Reproduce**
Start neolink with h264 stream and open with `ffplay`

Log will include items such as
```
[h264 @ 0x7f9866877c00] Invalid NAL unit 0, skipping.    0B f=0/0
[h264 @ 0x7f9866877c00] Invalid NAL unit 0, skipping.    0B f=0/0
[h264 @ 0x7f9866877c00] Invalid NAL unit 0, skipping.    0B f=0/0
[h264 @ 0x7f9866877c00] Invalid NAL unit 0, skipping.    0B f=0/0
[h264 @ 0x7f9866877c00] Invalid NAL unit 8, skipping.    0B f=0/0
```

**Expected behavior**
ffmpeg/ffplay to report a valid stream

**Versions**
Neolink version: head (95935d331b3c8b2de07a8d40ec7b9580a16424c7e)
Reolink camera model and firmware: E1

I think this is to do with the extra data that neolink sends along with the video such as the aac. Filtering this out may be sufficient.

QuantumEntangledAndy:
The data sent to gstreamer includes non video related data, such as audio. This causes some players to have to perform extra work to filter out invalid packets (see #49).

This PR uses the headers of each packet to deduce its types and the appropriate action.

fixes #49 

Proving that the same headers are used throughout the reolink family this should work for all, but it would be best to test on a few more cameras.


hh1209:
Would it be feasible to stream the "filtered out" audio to an RTSP audio stream?

QuantumEntangledAndy:
Yes that is my eventual plan. First step filter out from the video. Second step put it into own stream. I'm working on it, no guarantee though.

thirtythreeforty:
Nice sleuthing. If this is the case (and you are probably completely right) then the BC protocol has gone from "obnoxious" to "exceedingly terrible."  Why on earth wouldn't you make this two separate message types?

Would you mind explaining each of the magic values in a comment? Even just something like the following would be helpful:

```rust
/// Video I-Frame: signals start of H264/H265 I-Frame
```
etc.

I'm also curious about how you learned these. In particular the "binary packet with no header" is really annoying from an implementation standpoint, since we don't actually know what kind it is except "probably it goes with the last data." This "in band signaling" is problematic for correctness in case a continuation packet happens to have the same magic bytes as a new packet. Is there any indication of how much data goes with each magic value - kind of a "sub packet"?

It's okay if you don't know the answers - that's part of reverse engineering :)

Regarding the implementation, it might be better if the byte parsing was moved to the `bc` module. Maybe `ModernMsg` should not contain `Vec<u8>` but instead contain an `enum` of `VideoData`, `AudioData`, etc:

```rust
pub enum BinaryData {
    VideoData(Vec<u8>),
    AudioData(Vec<u8>),
    // do we need Unknown(Vec<u8>) here?
}

pub struct ModernMsg {
    pub xml: Option<BcXml>,
    pub binary: Option<BinaryData>,
}
```

and then `bc_protocol` can simply match on the type.

QuantumEntangledAndy:
The magic codes are based of the work of @twistedddx in issue #4. The format he described matached my own investigations and I worked it up into this PR.

With regards to extra info. After each magic header there is data such as the number of expected bytes in the packet. I believe that if this number is >40000 it gets chunked. We could set it up to expect this chunking but it seems like a lot of extra work.

After the magick header in the video stream there is also the text H264. We may be able to use this to auto detect the format.

QuantumEntangledAndy:
With regards to understandig why no header is a continued packet:
I had the program only print out unknown headers and looked for patterns. It was mostly random without a pattern indicating that it was just data. I then made it print know, unknown and size (its current print statements) and it became clear that this unknown was large packets that were chunked after the video.

Also if I told it to not parse the unknown headers to the video stream and tried to play it only the top half of the video was valid indicating that the later half of the frame was missing.

QuantumEntangledAndy:
Perhaps this will ultimately require a more complex structure for each type to parse correctly.

E.g. In video type We extract the data such as video frame size and type. I also think that the PTS and DTS are in the header so we could maybe attach proper time stamps.

QuantumEntangledAndy:
It is probably good for us that they don't use separate header for continuation.

I believe most players search the stream for the nal unit separators. [info on h264 stream](http://gentlelogic.blogspot.com/2011/11/exploring-h264-part-2-h264-bitstream.html?m=1)

This is why passing all the data is currently working. Our VLC and ffmpegs are scanning the streaming for the NAL start codes then playing assuming all the rest (until the next NAL unit start) is the video data. If there was extra data mid NAL unit we would probably be seeing artifacts in the video.

QuantumEntangledAndy:
I've gotten the basics for putting it into the enum you mentioned. Still have to refractor more if we want to properly handle that overflow issue you mentioned

twistedddx:
Something that may be worth knowing before it costs you hours of time:
The I and P frames listed data size is up to 16 bytes short of the actual data.
Eg header says 234567 bytes but then I found 234574 bytes of actual data followed.
If you stop writing at 234567 bytes you can get video distortion at the bottom of effected frames.
I observe this in the temp file the official client writes to disk when exporting.

Now that I see these chunked packets during transport maybe the difference is that the official client incorrectly picks up some bytes each time the data was chunked? Maybe there is a continuation byte marker that the client doesn't remove from the video frame.

QuantumEntangledAndy:
Thanks you for the info I will keep that in mind when I get around to keeping track of the sizes. But at it's currently implementation it is lazy and dosent worry about it. If there is a continuation marker maybe we are getting small artifacts... Hmmm

QuantumEntangledAndy:
I'm wondering if filtering all this out will help improve the UDP connections.

QuantumEntangledAndy:
@thirtythreeforty could you help level up my rust here:

My current implementation of parsing more structure from the video data looks like this

```rust
#[derive(Debug, PartialEq, Eq)]
pub struct VideoIFrame {
    pub magic: Vec<u8>,
    pub video_type: Vec<u8>,
    pub data_size: Vec<u8>,
    pub unknowna: Vec<u8>,
    pub timestamp: Vec<u8>,
    pub unknownb: Vec<u8>,
    pub clocktime: Vec<u8>,
    pub unknownc: Vec<u8>,
    pub video_data: Vec<u8>,
}

impl VideoIFrame {
    pub fn from_binary(binary: &[u8]) -> VideoIFrame {
        VideoIFrame {
            magic: binary[0..4].to_vec(),
            video_type: binary[4..8].to_vec(),
            data_size: binary[8..12].to_vec(),
            unknowna: binary[12..16].to_vec(),
            timestamp: binary[16..20].to_vec(),
            unknownb: binary[20..24].to_vec(),
            clocktime: binary[24..28].to_vec(),
            unknownc: binary[28..32].to_vec(),
            video_data: binary[32..].to_vec(),
        }
    }

    pub fn len(&self) -> usize {
        self.video_data.len() + 32
    }

    pub fn as_slice(&self) -> &[u8] {
        let mut result = self.magic.clone();
        result.extend(&self.video_type);
        result.extend(&self.data_size);
        result.extend(&self.unknowna);
        result.extend(&self.timestamp);
        result.extend(&self.unknownb);
        result.extend(&self.clocktime);
        result.extend(&self.unknownc);
        result.extend(&self.video_data);
        result.as_slice()
    }
}

#[derive(Debug, PartialEq, Eq)]
pub struct VideoPFrame {
    pub magic: Vec<u8>,
    pub video_type: Vec<u8>,
    pub data_size: Vec<u8>,
    pub unknowna: Vec<u8>,
    pub timestamp: Vec<u8>,
    pub unknownb: Vec<u8>,
    pub video_data: Vec<u8>,
}

impl VideoPFrame {
    pub fn from_binary(binary: &[u8]) -> VideoPFrame {
        VideoPFrame {
            magic: binary[0..4].to_vec(),
            video_type: binary[4..8].to_vec(),
            data_size: binary[8..12].to_vec(),
            unknowna: binary[12..16].to_vec(),
            timestamp: binary[16..20].to_vec(),
            unknownb: binary[20..24].to_vec(),
            video_data: binary[24..].to_vec(),
        }
    }

    pub fn len(&self) -> usize {
        self.video_data.len() + 24
    }

    pub fn as_slice(&self) -> &[u8] {
        let mut result = self.magic.clone();
        result.extend(&self.video_type);
        result.extend(&self.data_size);
        result.extend(&self.unknowna);
        result.extend(&self.timestamp);
        result.extend(&self.unknownb);
        result.extend(&self.video_data);
        result.as_slice()
    }
}

#[derive(Debug, PartialEq, Eq)]
pub enum VideoFrame {
    IFrame(VideoIFrame),
    PFrame(VideoPFrame),
}

impl VideoFrame {
    pub fn len(&self) -> usize {
        match self {
            VideoFrame::IFrame(binary) => binary.len(),
            VideoFrame::PFrame(binary) => binary.len(),
        }
    }

    pub fn as_slice(&self) -> &[u8] {
        match self {
            VideoFrame::IFrame(binary) => binary.as_slice(),
            VideoFrame::PFrame(binary) => binary.as_slice(),
        }
    }
}

#[derive(Debug, PartialEq, Eq)]
pub enum BinaryData {
    VideoData(VideoFrame),
    AudioData(Vec<u8>),
    InfoData(Vec<u8>),
    Unknown(Vec<u8>),
}
```

The trouble I am having is with the as_slice (used as part of the serialisation and the passing to gstreamer). I think this function is meant to represent the data a single continuous piece, but when the struct is constructed I break it up. In as_slice I try to reconstruct it as a single piece using extends. But it complains about how I am returning a loacally variable as a borrow

```
returns a value referencing data owned by the current function
```

I am thinking that I could instead write it as something like this:

```rust
pub struct VideoPFrame {
    pub binary Vec<u8>,
}

impl VideoPFrame {
    pub fn from_binary(binary: &[u8]) -> VideoPFrame {
        VideoPFrame {
            binary: binary.to_vec(),
        }
    }
   pub fn magic(&self) -> &[u8] {
      self.binary[0..4]
   }
   pub fn video_type(&self) -> &[u8] {
      self.binary[4..8]
   }
```

Instead but it doesn't feel as elegent. I suppose on the plus side though we could make:

```rust
enum VideoType {
   H264,
   H265,
   Unknown,
}
pub fn video_type(&self) -> VideoType {
      const h264_magic: &[u8] = &[0x48, 0x32, 0x36, 0x34];
      const h265_magic: &[u8] = &[0x48, 0x32, 0x36, 0x35];
      match self.binary[4..8] {
         h264_magic => VideoType::H264,
         h265_magic => VideoType::H265,
         _ => VideoType::Unkown,
      }
   }
```

What do you think would be best?

QuantumEntangledAndy:
I'll push my work so far and you can see it in the wild but it will fail to compile

QuantumEntangledAndy:
So I decided to go with the getters method where the struct just stores it as raw and then there are methods to get the data from it.

I am now having fun trying to get the correct size info, there seems to be some extra bytes somewhere when the frame is split

QuantumEntangledAndy:
I am having a sinking feeling about the extra bytes. The h264 format should have a start code of `0, 0, 1` the bc seems to have `0, 0, 0, 1`. I am wondering if it is null terminating things.... I really hope not but testing now

QuantumEntangledAndy:
Ok on the plus side it seems unrelated to the number of start code... on the negative side why are the start codes having an extra 0?

QuantumEntangledAndy:
So the number of extra bytes in the first frame seems to be arbitary I have seen 0, 3, 5 and 6 extra bytes without a pattern

Ok this seems a bit dodgy but.....

The frames are split like this:
- Full data 45000, chunked in 40000:
- first frame: 50000
- last frame: 400000

Example of weird count:
- Expected bytes 54843
- Bytes in first frame: 14848
- Bytes in second frame: 40000
- Total received bytes: 54848
- Difference: 5

In otherwords the first frame has the odd number of bytes

I am wondering if their rounding logic for the count is incorrect somehow.

I could get it to work by....

- Stopping when we get a negative number
- Use this to get correct bytes: expected_bytes - modulo(expected, 40000) + bytes_in_first_frame

QuantumEntangledAndy:
So I am having some success with the altered expected bytes `expected_bytes - modulo(expected, 40000) + bytes_in_first_frame` however everynow and then my trace messages seem to come out of order and then we get a panic because there seemed to be an audio frame mid chunked data. Are the messages from the camera coming in from some async source?

QuantumEntangledAndy:
So at this point I'm feeling that BC will either:

- drop the rest of the frame of it detects any of the other magic bytes. Even if it's not finished

- assume only video is chunked and will process audio while waiting for next non magic code to fulfill the frame

QuantumEntangledAndy:
So I have observed the start of another video frame even before the rest of the chucks are processed. So it seems to have dropped the packet for some reason.... Sigh

twistedddx:
More things I have observed in files written to disk.
The files are coming from a NVR with 8 cameras camera's connected. I have had SWN-NHD805/806/818, RLC-410-4MP/5MP and RLC-520 cameras connected. Obviously Neolink is more interested in data coming direct from cameras rather than from NVR.
Swann 805 and 806 cameras are no longer connected but are available for testing.
Swann cameras use the ADPCM audio, Reolink use AAC. This appears to just be a change in firmware that Swann never released for their products.

Several frames from another camera can appear mid stream.
Eg Camera 5 data is exported and Camera 2 data for a few frames can appear mid file. I saw this maybe 3 times within 100 test files. I did have old and "unsupported" cameras connected which may have caused additional problems.
Different cameras types leave different finger prints in the h264 data because of different encode settings eg width, height, bitrate, profile etc.
For P frames byte 22 was significant and for I frames byte 5 was significant. If that byte changed, the data had come from a different model camera and I drop the data.

P or I frames may end prematurely. This sounds like your described BC stop processing a I or P frame if it is interrupted by the start of another frame.

Audio is not always a single frame at a time, but extremely often is.

BC pad the frames with zeros to byte align to the next 8 byte boundary. But rarely a frame does not start on a 8 byte boundary.

Basically there are no rules and data corruption is not uncommon.

thirtythreeforty:
> ```rust
> impl VideoIFrame {
>     pub fn as_slice(&self) -> &[u8] {
>         let mut result = self.magic.clone();
>         result.extend(&self.video_type);
>         result.extend(&self.data_size);
>         result.extend(&self.unknowna);
>         result.extend(&self.timestamp);
>         result.extend(&self.unknownb);
>         result.extend(&self.clocktime);
>         result.extend(&self.unknownc);
>         result.extend(&self.video_data);
>         result.as_slice()
>     }
> }
> ```
> 
> The trouble I am having is with the as_slice (used as part of the serialisation and the passing to gstreamer). I think this function is meant to represent the data a single continuous piece, but when the struct is constructed I break it up. In as_slice I try to reconstruct it as a single piece using extends. But it complains about how I am returning a loacally variable as a borrow
> 

Right - this is basically stopping you from returning a dangling pointer. To do this you would need to return a `Vec<u8>`, the "owned" version of a byte slice.

> I am thinking that I could instead write it as something like this:
> 
> ```rust
> pub struct VideoPFrame {
>     pub binary Vec<u8>,
> }
> 
> impl VideoPFrame {
>     pub fn from_binary(binary: &[u8]) -> VideoPFrame {
>         VideoPFrame {
>             binary: binary.to_vec(),
>         }
>     }
>    pub fn magic(&self) -> &[u8] {
>       self.binary[0..4]
>    }
>    pub fn video_type(&self) -> &[u8] {
>       self.binary[4..8]
>    }
> ```
> 
> Instead but it doesn't feel as elegent. What do you think would be best?

I would say the latter is more elegant. You're returning a reference with a lifetime based on the original argument's lifetime (a reference to borrowed data).  Without lifetime elision, this looks like:

```rust
pub fn magic<'a>(&'a self) -> &'a [u8] {
    self.binary[0..4]
}
```

-- here `'a` is the lifetime of `self`.

All this said, however, it looks like you're trending further toward writing a real parser, for which I'm currently using Nom. Although I'm not super happy with it, it does give you a couple niceties, like being able to operate directly on a stream. The deal with Nom is that you return functions ("combinators") that do the actual parsing. This lets the Nom library control parsing execution.

Don't worry too much about Nom for now, just wanted to put the bug in your ear. I'd say let's figure out the BC data stream format first, then we can worry about the implementation.  I'll put more comments about the format in a second comment since this one is getting long...

QuantumEntangledAndy:
Yes I'm thinking prematurely ending frames is my issue. I have to experiment more to see if it will send the data anyways at some later time. If it does this will be a nightmare to process. If for example I have two large frames both sending packets at the before the other is finished their dosent seem to be a way to untangle them.

I'm thinking to do this:

- use my corrected byte size and wait for the bytes. (Ideal case)
- if a magic code comes along abandon the waiting for packets and just process that (assume frame will never be finished)
- If we are expecting a continuation packet but it does not arrive as a known magick code and is not of size 40000 stop processing frame and log it first few bytes (assume it is an unknown magic code and assume frame will never be finished)
- If we are not expecting a continuation packet and we get a packet with an unknown magic and of size 40000 then treat it as one of the lost continuation packets (drop it but log it as it may help understand the format more)

Will report back on what this yields. 

P.s. I think I saw a iframe with only 24 bytes total at one point.... So not even a full header.

thirtythreeforty:
> So the number of extra bytes in the packet frame seems to be arbitary I have seen 0, 3, 4, 5 and 6 extra bytes without a pattern. This number of extra bytes does not scale linearly with the number of chunks indicating not a continuation marker.
> 
> Example of weird count:
> 
>     * Expected bytes 54843
> 
>     * Bytes in first packet: 14848
> 
>     * Bytes in second packet: 40000
> 
>     * Total received bytes: 54848
> 
>     * Difference: 5
> 
> 
> In otherwords the first packet has the odd number of bytes

I am seeing something similar in a packet capture and doing the math by hand with Wireshark helping.

Screenshot_20200730_090052

My Baichuan plugin doesn't understand the stateful `binary` so ignore the yellow.  The counts are correct on the subsequent packets, so the binary payload is:

- 30312 = 30344 - 32 byte header
- 280000 = 40000 * 7

for a sum of 310312 bytes. The binary header format matches your code and spec (bravo!) and claims a payload length of 0x04bc23 = 310307.  A difference of 5 extra bytes sent.

> I am wondering if their rounding logic for the count is incorrect somehow.

Maybe. However, if that were the case I think you would not see the behavior for a single-message video packet, which I happen to have an example of (H264 substream I-Frame, again from B800):

Screenshot_20200730_091116

You can see from the sizes that the video info fits in a single BC packet. Now, the binary payload size is simply:

- 10728 = 10760 - 32 byte header

and their claimed payload length is 0x29e3 = 10723, a difference of 5 bytes *in this single packet*.

Here's the packet if you want to look at it: [h264-iframe-wrong-claimed-packets.pcapng.gz](https://github.com/thirtythreeforty/neolink/files/5001725/h264-iframe-wrong-claimed-packets.pcapng.gz)

> 
> I could get it to work by....
> 
>     * Stopping when we get a negative number
> 
>     * Use this to get correct bytes: expected_bytes - modulo(expected, 40000) + bytes_in_first_packet
> 
> 
> I have been looking at the beginning and end of each packet, with and without a header and there is no pattern that I can see. No numbers near the beginning or end of the packet correspond to anything like the packet size and the hexes are not ASCII either

As far as I can tell, they're not padding with garbage either - if that were the case, you might expect the end of the last buffer to equal the end of the previous buffer, but it doesn't.

thirtythreeforty:
@QuantumEntangledAndy can you post a sample video message that has 6 bytes extra or 4 bytes extra or whatever? All the examples we've discussed are 5 extra bytes. It would be interesting to see where the extra bytes appear.

thirtythreeforty:
> Something that may be worth knowing before it costs you hours of time:
> The I and P frames listed data size is up to 16 bytes short of the actual data.
> Eg header says 234567 bytes but then I found 234574 bytes of actual data followed.
> If you stop writing at 234567 bytes you can get video distortion at the bottom of effected frames.
> I observe this in the temp file the official client writes to disk when exporting.
> 
> Now that I see these chunked packets during transport maybe the difference is that the official client incorrectly picks up some bytes each time the data was chunked? Maybe there is a continuation byte marker that the client doesn't remove from the video frame.

I am tempted to just roll with it and say "yep, expect up to 16 extra bytes of data."  (Re: "costs you hours of time"!) I would not be surprised if this just boils down to an accounting bug that gets hidden from Reolink because they're using their own library to do the unparsing and it's not fussy about byte lengths.

twistedddx:
> I am tempted to just roll with it and say "yep, expect up to 16 extra bytes of data." (Re: "costs you hours of time"!) I would not be surprised if this just boils down to an accounting bug that gets hidden from Reolink because they're using their own library to do the unparsing and it's not fussy about byte lengths.

Yep once QuantumEntangledAndy found it was not a marker to note the packet is continuing I go back to my answer from last year of poor maths upstream.

thirtythreeforty:
Alright, let's do this. Roll with @twistedddx's 16 extra bytes.

@QuantumEntangledAndy - here's how I think data should flow:

1. Message comes in from wire. `bc::de` parses it into basically the structures you have now, with two important changes:
  - First, split the binary payload into header/video. I don't really think the current Neolink implementation will care much about the contents of the binary header, although you're correct that it could be used for very clean format autodetection.
  - Second, maintain a copy of the message's binaryData current frame type (Iframe, audio, whatever) in the `context`, similar to how it's done currently for checking whether the message ID is in binary mode...

https://github.com/thirtythreeforty/neolink/blob/5935d331b3c8b2de07a8d40ec7b9580a16424c7e/src/bc/de.rs#L119

...so that you can correctly emit "continuation" packets, whether they're from audio or video. This will need a new data structure in the context.  If you want to keep count of lengths, you can just use `<=` when checking, with a note about the suspected poor accounting.

2. Parsed message flows to bc_protocol which demuxes the binaryData into video/audio streams. The `de` API never gives raw copies of the header, only the body (as a `Vec<u8>`).

src/bc/model.rs
```diff
@@ -22,10 +23,184 @@ pub enum BcBody {
...
+    pub fn magic(&self) -> &[u8] {
+        &self.raw[0..4]
+    }
+
+    pub fn video_type(&self) -> &[u8] {
+        &self.raw[4..8]
+    }
+
+    pub fn data_size(&self) -> usize {
+        u32::from_be_bytes(self.raw[8..12].try_into().expect("slice with incorrect length")).try_into().expect("u32 won't fit into usize")
```
thirtythreeforty, line 118: You get this basically for free with Nom combinators

src/bc/model.rs
```diff
@@ -22,10 +23,184 @@ pub enum BcBody {
...
+        &self.raw[32..]
+    }
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct VideoPFrame {
+    pub raw: Vec<u8>
+}
+
+impl VideoPFrame {
```
thirtythreeforty, line 94: I think you should have a single `BinaryData` struct like the following:

```rust
pub struct BinaryData {
    data: Vec<u8>,
}

impl BinaryData {
    pub fn body(&self) -> &[u8] { /* ... */ }
    pub fn kind(&self) -> BinaryDataKind { /* match on header */ }
}
```

and don't bother with parsing the header much right now.  We can add helpers to give `bc_protocol` more info if it needs it, and this way you have a single struct to deal with, with none of the duplication that IFrame, PFrame causes.

thirtythreeforty: `bc_protocol` can `match` on the `BinaryDataKind`. This separates the "control path" from the "data path", keeping the latter using a single data type.

thirtythreeforty:
Your implementation before your last push kinda matches my description, although you would still need to keep track of the current `BinaryDataKind` in the context, which is sadly necessary work to make Neolink *correct* (an important quality of a security system!)

QuantumEntangledAndy:
Heres some examples of other incorrect sizes:

- 0 bytes
  - header: 30, 31, 64, 63, 48, 32, 36, 34, 60, 18, 0, 0, 0, 0, 0, 0, c8, b9, 2e, 3d, 49, 0, 0, 0
  - first bytes: 0, 0, 0, 1, 21, 9b, fc, 2f, ff, 4a, dd, 7a, 79, cc, a9, 38
  - last bytes: b4, 7d, 4f, 6d, 60, f, 2f, e, 73, 22, de, e1, a8, 8, 68, c0
  - expected bytes: 6240
  - actual bytes: 6240

- 2 bytes:
  - header: 30, 31, 64, 63, 48, 32, 36, 34, 86, 18, 0, 0, 0, 0, 0, 0, 8, c7, 31, 3d, 49, 0, 0, 0
  - first bytes: 0, 0, 0, 1, 21, 9a, 24, 2f, ff, 4a, dd, 7b, b6, 33, 8e, 39
  - last bytes: 68, 2a, 51, 1b, d1, 28, e7, e8, 80, 0, 0, 0, 0, 0, 0, 0
  - expected bytes: 6278
  - actual bytes: 6280

- 3 bytes
  - header: 30, 31, 64, 63, 48, 32, 36, 34, 1d, 19, 0, 0, 0, 0, 0, 0, 78, 9d, 27, 3d, 49, 0, 0, 0
  - first bytes: 0, 0, 0, 1, 21, 9b, 0, 2f, ff, 4a, dd, 17, db, 73, 15, cb
  - last bytes: d, b5, 2e, 75, 5d, 26, 9c, 88, a1, 2a, b5, 3b, c0, 0, 0, 0
  - expected bytes: 6429
  - actual bytes: 6432

- 4 bytes
  - header: 30, 31, 64, 63, 48, 32, 36, 34, 54, 18, 0, 0, 0, 0, 0, 0, 80, bf, 2f, 3d, 49, 0, 0, 0
  - first bytes: 0, 0, 0, 1, 21, 9a, 0, 2f, ff, 14, ad, 39, b0, 52, 7b, 83
  - last bytes: 1, fa, e1, 34, 8f, e6, 6b, c4, 12, 8b, 93, 88, 0, 0, 0, 0
  - expected bytes: 6228
  - actual bytes: 6232

- 6 bytes
  - header: 30, 31, 64, 63, 48, 32, 36, 34, a, 19, 0, 0, 0, 0, 0, 0, b8, aa, 2a, 3d, 49, 0, 0, 0
  - first bytes: 0, 0, 0, 1, 21, 9b, 6c, 2f, ff, 1, 48, 55, 58, 6c, 56, 47
  - last bytes: 55, 6e, 66, ed, 93, 45, 60, f7, f4, 8c, 0, 0, 0, 0, 0, 0
  - expected bytes: 6410
  - actual bytes: 6416

- 7 bytes
  - header: 30, 31, 64, 63, 48, 32, 36, 34, b9, 18, 0, 0, 0, 0, 0, 0, 50, c1, 30, 3d, 49, 0, 0, 0
  - first bytes: 0, 0, 0, 1, 21, 9a, 24, 2f, ff, 4a, dd, 7b, b6, 33, 8e, 39
  - last bytes: 68, 2a, 51, 1b, d1, 28, e7, e8, 80, 0, 0, 0, 0, 0, 0, 0
  - expected bytes: 6329
  - actual bytes: 6336

QuantumEntangledAndy:
Having it all in one place is actually helping here: We have zero terminated packets, this is without chuncking I will look into the chunked packets

thirtythreeforty:
Yep, the nice layout helps. I agree the zeros at the end are common. Possibly we need to throw those away?

QuantumEntangledAndy:
Yes I suspect lets see if this continues with the large chuncked frames

QuantumEntangledAndy:
Chunked frames:

- 6 bytes:
  - header: 30, 30, 64, 63, 48, 32, 36, 34, fa, 2b, 7, 0, 8, 0, 0, 0, 28, 7e, e2, 3c, 49, 0, 0, 0, 73, 77, 23, 5f, 22, 1, c3, 0
  - packets
    - 0
      - first bytes: 0, 0, 0, 1, 67, 64, 0, 33, ac, e8, 2, 40, a, 39, 0, 0
      - last bytes:
    - 
      - first bytes: 
      - last bytes:

    - 
      - first bytes: 
      - last bytes:

    - 
      - first bytes: 
      - last bytes:

    - 
      - first bytes: 
      - last bytes:

    - 
      - first bytes: 
      - last bytes:

QuantumEntangledAndy:
So dosen't seem to be null terminated with the chunked frames

QuantumEntangledAndy:
- 1 bytes:
  - header: 30, 30, 64, 63, 48, 32, 36, 34, d7, 72, 3, 0, 8, 0, 0, 0, 60, e9, 7b, aa, 49, 0, 0, 0, a2, 7e, 23, 5f, 22, 1, c3, 0
  - expected: 226007
  - actual: 226008
  - packets
    - 0
      - size: 26040
      - size excluding header: 26008
      - first bytes: 0, 0, 0, 1, 67, 64, 0, 33, ac, e8, 2, 40, a, 39, 0, 0
      - last bytes: 3b, 53, d6, 87, 9b, ac, 47, 64, 27, 4a, a3, f7, c7, 41, 16, 4
    - 1
      - size: 40000
      - first bytes: 32, ee, fe, ed, 5b, d1, ff, b5, 8e, 7e, b1, 8b, e, 42, 3d, c7
      - last bytes: 67, a5, c1, 7a, a, d3, c5, 5, 56, 5f, ab, d7, 51, dc, 6d, cf
    - 2
      - size: 40000
      - first bytes: f5, ab, 0, a5, 74, a5, b0, 8a, d3, 55, b0, 6c, a9, c8, 6, e3
      - last bytes: 2d, 12, 53, c6, 2e, 81, 7c, 26, b5, b9, 6e, ff, 29, c9, aa, 15
    - 3
      - size: 40000
      - first bytes: cc, 88, 90, 41, 60, c5, da, 2, 7e, ab, 7c, 9, 3a, a1, 95, bf
      - last bytes: d9, 6b, 11, 1, b0, 45, c4, 85, ae, 37, 7f, 75, 52, 22, 94, b3
    - 4
      - size: 40000
      - first bytes: fd, bc, f0, b8, d8, 94, ef, de, 2b, aa, 3d, 42, 23, 82, 23, ae
      - last bytes: 9, 51, d4, c8, 1f, 1f, 1f, b8, eb, b2, 7c, b7, 85, f1, 6, e7
    - 5
      - size: 40000
      - first bytes: 43, a8, bf, 8, 1a, c0, b0, d3, dd, 1d, c5, c8, ea, 47, 91, eb
      - last bytes: c9, bd, b, 1f, 7d, 81, bd, 39, fe, eb, 7, b4, d6, 7f, e0, a2


- 4 bytes:
  - header: 30, 30, 64, 63, 48, 32, 36, 34, 6c, 9, 3, 0, 8, 0, 0, 0, f8, 69, 9a, aa, 49, 0, 0, 0, a4, 7e, 23, 5f, 22, 1, c3, 0
  - expected: 199020
  - actual: 199024
  - packets
    - 0
      - size: 39056
      - size excluding header: 39024
      - first bytes: 0, 0, 0, 1, 67, 64, 0, 33, ac, e8, 2, 40, a, 39, 0, 0
      - last bytes: b1, 96, 24, 58, 7a, 6e, f8, c2, a5, 68, 6e, f4, 7a, 1e, 4f, af
    - 1
      - size: 40000
      - first bytes: 21, 52, 22, 54, bd, c2, c8, 30, 1c, 34, 36, d7, ab, 1e, 68, f1
      - last bytes: 79, 7d, f1, a4, 22, ad, bd, b1, a0, 60, 65, 37, 3b, e0, 48, 1e
    - 2
      - size: 40000
      - first bytes: 65, 5a, b, ff, 9d, db, e8, 2f, d6, 3a, 7a, f4, d8, 52, ab, 33
      - last bytes: 8d, a2, 6f, 5, f9, 30, 4f, bc, 62, a9, 3, b4, 88, b4, ef, ba
    - 3
      - size: 40000
      - first bytes: 59, 61, 9d, 10, 51, 18, 9b, d, c6, cd, b5, a, 5f, dc, ba, c1
      - last bytes: 87, c3, 93, 4f, 67, 78, a6, 21, f5, 46, 46, 9f, c4, 89, 95, f8
    - 4
      - size: 40000
      - first bytes: c9, 75, 3, cf, 82, 29, ae, 15, e, f0, bb, ba, e3, 55, 4b, 14
      - last bytes: 57, 9f, 39, b4, de, aa, b, a3, fb, 3c, b1, fe, a4, 7e, 23, 5f

QuantumEntangledAndy:
I am wondering if we should do the same and throw away the last few bytes of the first frame even though they aren't zeros. Perhaps BC is being lazy and just leaving them as random data

twistedddx:
I dont recall if I ever saw the extra bytes as larger than 7. As I did not understand the problem I may have just decided up to 16 will surely never fail me. 

It very much looks like the actual bytes is always a multiple of 8 meaning difference between expected and actual could be 0 to 7.

QuantumEntangledAndy:
Yes we seem to have 8 byte padding that is zero in single packets and random data in contiuations. This must have been a nightmare for your ffmpeg to handle

QuantumEntangledAndy:
So based on the assumption that extra bytes are junk I have refractored this in the style your comments mentioed. 

I however do not know how to emit the contiuation you mentioed

src/bc_protocol.rs
```diff
@@ -242,7 +242,7 @@ impl BcCamera {
         Ok(())
     }
 
-    pub fn start_video(&self, data_out: &mut dyn Write, stream_name: &str) -> Result<Never> {
+    pub fn start_video(&mut self, data_out: &mut dyn Write, stream_name: &str) -> Result<Never> {
```
QuantumEntangledAndy, line 245: Revert mutable, since I don't need that anymore

QuantumEntangledAndy: Done

QuantumEntangledAndy:
I think I get what you mean about the context, I didn't realise that was your object holding info on previous frames, but I traced the function up and worked it out.

Please have a look at my continuation model

thirtythreeforty:
Yeah, whether `binaryData` is present at all is actually stateful, which is already stored in the context. Let me review your new implementation.

src/bc/model.rs
```diff
@@ -22,10 +26,154 @@ pub enum BcBody {
...
+}
+
+impl BinaryData {
+    pub fn body(&self) -> &[u8] {
+        let lower_limit = self.header_size();
+        let upper_limit = self.body_size() + lower_limit;
+        &self.data[lower_limit..upper_limit]
+    }
+
+    pub fn header_size(&self) -> usize {
```
thirtythreeforty, line 55: Oh ffs Reolink, really? A separate header size per kind?

QuantumEntangledAndy:
Sorry about the rustfmt getting in the way of things I am trying to get into the habit of using it, but really should have done it as a last commit when we are all done

src/bc/model.rs
```diff
@@ -65,6 +213,8 @@ pub(super) struct BcSendInfo {
 #[derive(Debug)]
 pub struct BcContext {
     pub(super) in_bin_mode: HashSet<u32>,
+    pub(super) last_binary_kind: Option<BinaryDataKind>,
+    pub(super) remaining_binary_bytes: usize,
```
thirtythreeforty, line 217: So everything looks great, except right here there is a subtle issue: the parsing context is actually per-connection, not per-message ID. So with this implementation there would be a problem if there were multiple message IDs with binaryData coming in - they would both overwrite this field.

That's why `in_bin_mode` is a HashSet: I check for the presence/absence of the message ID in that set.

You should do the same thing here, with a HashMap to store your two new fields *per message ID*. If you were more adventurous, you could actually combine `in_bin_mode` with your new HashMap, since their meaning is identical.

thirtythreeforty:
> Sorry about the rustfmt getting in the way of things I am trying to get into the habit of using it, but really should have done it as a last commit when we are all done

No worries, those snuck in because I didn't run it before commit.

src/bc/model.rs
QuantumEntangledAndy: Yes, sorry

src/bc/model.rs
QuantumEntangledAndy: I haven't look at the msgid but will they be the same for the same set of continuations?

thirtythreeforty: Yes, all the video msgids are 3, for example. In practice that's the only binary-enabled message ID I've seen so far, so this hashmap will normally only have one member.

QuantumEntangledAndy: Ahh yes I recall looking into that when I first started this branch ok! I am on it

QuantumEntangledAndy: Also last_binary_kind and in_bin_mode are not quite the same since I leave it as None if I don't expect a continuation

thirtythreeforty: Fair point. I meant more, the presence of your new struct, at all, in the HashMap, should always be the same as the HashSet, making the Set redundant.

thirtythreeforty: But since this is not really a performance issue, I am not picky either way.

src/bc/model.rs
thirtythreeforty: Not your fault! I was angry at reolink haha.

QuantumEntangledAndy:
So now using hashmaps please have another look

QuantumEntangledAndy:
So I am running this through its excersize now and I got an unexpected situation. Header suggests size is 16392 (not big enough for chunking) but length is 5328 so the `binary_data.body()` fails... I will have to add upper bounds on the slice index

QuantumEntangledAndy:
So I have gotten the upper limit set but ffprobe is suggesting I have invalid IFrames something about missing left bits

QuantumEntangledAndy:
It will play but there are a lot of artifacts

QuantumEntangledAndy:
Ok I dropped the incomplete packets and now there are less artifacts but still some. I will keep going

QuantumEntangledAndy:
I get a completly valid stream with this:


```rust
match binary.kind() {
                        BinaryDataKind::VideoDataIframe
                        | BinaryDataKind::VideoDataPframe
                        | BinaryDataKind::VideoCont => {
                            data_out.write_all(&binary.data)?;
                        }
                        BinaryDataKind::AudioDataAac
                        | BinaryDataKind::AudioDataAdpcm
                        | BinaryDataKind::AudioCont => {
                            ();
                        }
                        BinaryDataKind::InfoData | BinaryDataKind::InfoDataCont => {
                            ();
                        }
                        BinaryDataKind::Unknown => {
                            data_out.write_all(&binary.data)?;
                        }
                    };
```

But now with this:

```rust
match binary.kind() {
                        BinaryDataKind::VideoDataIframe
                        | BinaryDataKind::VideoDataPframe
                        | BinaryDataKind::VideoCont => {
                            data_out.write_all(binary.body())?;
                        }
                        BinaryDataKind::AudioDataAac
                        | BinaryDataKind::AudioDataAdpcm
                        | BinaryDataKind::AudioCont => {
                            ();
                        }
                        BinaryDataKind::InfoData | BinaryDataKind::InfoDataCont => {
                            ();
                        }
                        BinaryDataKind::Unknown => {
                            data_out.write_all(binary.body())?;
                        }
                    };
```

This suggests that I am dropping some part of the packet we shouldn't and also not sending the completed frames during continuation

QuantumEntangledAndy:
Also invalid:

```rust
match binary.kind() {
                        BinaryDataKind::VideoDataIframe
                        | BinaryDataKind::VideoDataPframe
                        | BinaryDataKind::VideoCont => {
                            data_out.write_all(binary.body())?;
                        }
                        BinaryDataKind::AudioDataAac
                        | BinaryDataKind::AudioDataAdpcm
                        | BinaryDataKind::AudioCont => {
                            ();
                        }
                        BinaryDataKind::InfoData | BinaryDataKind::InfoDataCont => {
                            ();
                        }
                        BinaryDataKind::Unknown => {
                            data_out.write_all(&binary.data)?;
                        }
                    };
```

QuantumEntangledAndy:
If I make body include the trailing bytes we drop it works. I'm thinking I need drop the last few bytes of the last ccontinuation packet. rather then those in first packet. This will not be fun to keep track of.

QuantumEntangledAndy:
I won't have time to finish this up right now but I wanted to say that they picture quality improves a lot. It is free of artifacts right from the start and the frames are updating cleanly when a fast object crosses the field.

thirtythreeforty:
Totally fine, I will run with it tonight (I'm in CST). Thanks for the research!

QuantumEntangledAndy:
I see if your CST and I'm UTC+7 that's a difference of 13 hours. I had a feeling it was something like this. 

Anyways I'm planning to do a refractor on the continuation logic trying to making it cleaner while I'm at it.

thirtythreeforty:
I have a rough implementation that discards the trailing bytes at the end of the continuations.  I'll push it soon as I clean it up.

thirtythreeforty:
Ok that took longer than I thought. It is pushed to the `magic_packet` branch in my repo. Butter smooth on my B800! 

A few things, please comment. 
- I am formalizing the inner packet structure; it's now termed a "media packet" (distinct from the outer "Baichuan packet").
- I removed the `Cont` variants from the kind type, since bc_protocol doesn't really care if it's a continuation or not.
- I keep track of the total reported length, subtracting as packets are received
- I slurped your code from model.rs into de.rs

I'm excited to add audio now!

thirtythreeforty:
Gah, spoke too soon. Just got a panic,
```
thread '<unnamed>' panicked at 'index 12 out of range for slice of length 8', src/bc/de.rs:312:43
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I know the problem - we're slicing everywhere, and we ought to either do a length test, or use .get(). I'll do the length check.

QuantumEntangledAndy:
I see, I've almost finished my version too haha. I have a function in mine called `complete` and I don't send the packets until I get all the data.

thirtythreeforty:
Ah I misunderstood, I thought you were taking a break from it. Didn't mean to step on each others' toes.

So, you're caching the incomplete media packet until you receive the entire length they claim?

QuantumEntangledAndy:
Was planning to take a break, but couldn't quite put it down haha. But kinda caching yes...

It works like this

Declare packet types we want 

Loop until we find a packet of that type

while ! packet.complete() {
   get next_packet
   if cont add next_paket to packet
   if another packet of interesting type make that the packet and try to finish it
  anything else restart proccess
}

// Now have complete packet
transmit

QuantumEntangledAndy:
I reverted de to your original as I feel like a packet desearliser should not care about the continuation it should just say whats there

QuantumEntangledAndy:
I am keeping de free from cont and instead adding it to the BcSubscription so that instead of calling `sub_video.rx.recv_timeout(rx_timeout)?;` we call `sub_video.get_binary_data_of_kind([list_of_kinds])?;`

QuantumEntangledAndy:
Here it is, have a look and maybe we can take the best of both yours and mine

QuantumEntangledAndy:
I'm trying to understand your version but what happens in this case:

Got unfinished video packet
Got audio packet 
Got cont (from video)

Have we already sent the video packets first part and we won't send the rest?

I had issues with doing that with errors popping up in ffplay. So I swapped to my current method.

QuantumEntangledAndy:
Also I don't mind if you want to run with yours. I am quite excited to tackle audio too

thirtythreeforty:
> I'm trying to understand your version but what happens in this case:
> 
> Got unfinished video packet
> Got audio packet
> Got cont (from video)

I had not anticipated continuations - in principle there's no way for us to figure out what the continuation is a part of in that case.  For your specific example, my branch would:

- Detect the incomplete video data, forward it
- Detect the audio data, forward it
- Forward the cont as "unknown" since it wasn't expecting it (audio packet erases the state from video packet)

@twistedddx did you ever observe an audio packet dropped in the middle of a video packet in the way Andy is describing? Like I said, there's no way to handle that general case because we won't be able to figure out where one continuation-kind ends and the other begins.

thirtythreeforty:
I just enabled audio on my camera (for some reason it was off) and I'm definitely seeing plopped audio packets.

This is truly awful. I would really like to know how this works in the official Reolink client; they must have some way of doing demux.

thirtythreeforty:
Ok, I've figured out exactly what is causing so much trouble. The embedded media packets form their own stream that does not necessarily align with the BC packet boundaries.  I have a sample BC packet that contains two media packet headers, one at the top as we know how to decode, and another in the middle.  The first packet is AAC, the second is the start of a video Pframe.  So, tossing these "extra bytes" would both lose the top of the Pframe, *and* cause an Unknown frame after it.

This isn't necessarily insane on Reolink's part - it's an embedded stream. But, at this point it really needs its own parsing layer kinda like you're already trending toward, Andy.

QuantumEntangledAndy:
I see. Then we need to track current packet and size. Skip that size in the stream and check the next one....

Can you send the example packet. I want to check if it's still at least 8byte aligned.

I'm thinking this

- Create buffer
- Check header at top.
- If have enough outer-packets in buffer for full size
  - Pop inner-packet
- else
  - Get extra outer-packets until we do
- If inner-packet is kind of interrst
  - Return inner-packet and stop
  - Retain buffer for next call
- Else drop inner-packet and keep going

QuantumEntangledAndy:
Alternatively if it's 8byte aligned we could just stride 8 bytes looking for magic codes.

twistedddx:
I see this BC media format outside the BC network protocol.
I believe the client writes all the media payload data to file with no check at that point.
What pain it goes through from there I do not know.

Almost all data written to file is 8 byte aligned, except in rare cases, I believe always after a prematurely ended previous frame.

00dc and 01dc etc can appear in what I believe is completely valid h264 data.
The magic sequence is just too short to brute force check for inside endless data.

My horrible but mostly working approach was:
1)Read a chunk of file larger than an I frame is expected to be and scan for 0x30 (start of BC media format frame).
2)then check byte sequences for video or audio.

magic sequences:
byte[] byte00dcH264Sequence = new byte[] { 0x30, 0x30, 0x64, 0x63, 0x48, 0x32, 0x36, 0x34 };  //00dcH264
byte[] byte01dcH264Sequence = new byte[] { 0x30, 0x31, 0x64, 0x63, 0x48, 0x32, 0x36, 0x34 };  //01dcH264
byte[] byteVOXSequence = new byte[] { 0x30, 0x31, 0x77, 0x62, 0xF8, 0x00, 0xF8, 0x00, 0x00, 0x01 }; //01wb [4]-[7] are typical size 248
byte[] byteAACSequence = new byte[] { 0x30, 0x35, 0x77, 0x62, 0x07, 0x02, 0x07, 0x02, 0xFF, 0xF1 };

Video 0-7 is tested, audio 0-3 + 8-9 is tested. Audio data size is normally the same size but not always, so cannot be tested.
These sequences were long enough to not get false positive matches.

3)Go find the byte sequence at the end that starts a new frame.
4)Write data between the magic sequences.

QuantumEntangledAndy:
I'm thinking to go with my above first implementation and at first assume it won't intruupt a packet.

In this model all packets should be of known type.

If I hit an unknown type I will assume that the camera had an error and intrupted a packet and that I should go into an error recovery mode by scanning for start codes like you suggested.

QuantumEntangledAndy:
My rust foo needs an upgrade. Can anyne explain this to me...

I am trying to seperate BinaryData out of model.rs into its own rust file media_packet.rs. But when I change hte imports I get this

```rust
error[E0432]: unresolved import `crate::bc::media_packet`
 --> src/bc_protocol.rs:4:16
  |
4 | use crate::bc::media_packet::*;
  |                ^^^^^^^^^^^^ could not find `media_packet` in `bc`

```

Do I need to do something to declare that media packet is part of bc? My googli is falling short but I'll keep looking

thirtythreeforty:
>
> My rust foo needs an upgrade. Can anyne explain this to me...
>
> I am trying to seperate BinaryData out of model.rs into its own rust file
> media_packet.rs. But when I change hte imports I get this
>
> error[E0432]: unresolved import `crate::bc::media_packet`
>  --> src/bc_protocol.rs:4:16
>   |4 | use crate::bc::media_packet::*;
>   |                ^^^^^^^^^^^^ could not find `media_packet` in `bc`
>
> Do I need to do something to declare that media packet is part of bc? My
> googli is falling short but I'll keep looking
>
Yes, declare it as a sub module in mod.rs:

    mod media_packet;

>

QuantumEntangledAndy:
Thanks I shall consider my rust foo as Lvl+1

QuantumEntangledAndy:
Ok made my head hurt but I've gotten it implemented please have a look and really stress test it

QuantumEntangledAndy:
Works like this:

- Call `get_media_packet_of_kind` with list of intereseted kinds
  - This function calls `next_media_packet` until it finds one of interest


- `next_media_packet` works as:
  - Advance the queue until we get a magic byte sequence
    - pulling more data as needed from the camera
    - Drop the data from the queue until we reach a magic byte
  -  Get the header of full header for that type of data
    - Pull data from the camera as needed
  - Get the data_size and pad_size from the header
  - Pull data from the camera to fill queue up to header + data + pad size
  - Pop the full media packet + pad from the queue
  - Make a MediaData struct from the data and return

This actually leaves the bc message as `Vec[u8]` again. Which I like as it separates arbitrary binary data from the data of video type (msgid 3).

src/bc_protocol.rs
```diff
@@ -10,7 +12,7 @@ use std::time::Duration;
 
 use Md5Trunc::*;
 
-mod connection;
+pub mod connection;
```
QuantumEntangledAndy, line 15: I changed this to pub so that I could access it from MediaData but thinking about it now I could: Move media_packet.rs into bc_protocl/ and then I don't need this.

Would you prefer that?

src/bc_protocol.rs
```diff
@@ -266,20 +271,48 @@ impl BcCamera {
...
+                }
+                MediaDataKind::AudioDataAac | MediaDataKind::AudioDataAdpcm => {
+                    // We should not have a packet of this kind as we skip them until we get
+                    // a packet of interest
+                    unreachable!();
+                }
+                MediaDataKind::InfoData => {
+                    // We should not have a packet of this kind as we skip them until we get
+                    // a packet of interest
+                    unreachable!();
```
QuantumEntangledAndy, line 298: I have these as unreachable at the moment mostly to ensure that I don't get unexpected data sneaking though during testing. When we've stress tested it enough I'd like to remove them.

QuantumEntangledAndy:
I am hoping this fixes the gray screen issue in #46 but if not I am wondering if there's something more in the H265 streams. Maybe another magick header for the new format. If I get the chance I will peek at your sample packet dumps.

QuantumEntangledAndy:
Umm quick question, if your free, about your samples in `src/bc/samples` they seem to be just binary dumps, which is OK as I can parse them to hex alright, but are they still Charlie scrambled? Do they include the xml header or are they just the binary part?

QuantumEntangledAndy:
Oh I think I have it sorted out `modern_video_start2.bin` is charlie un-scrambled dumps of the binary part with some sort of extra header which I think is part of your ethernet frame, the actual data starts 24 bytes in

QuantumEntangledAndy:
I can't see anything new in the relevent dump. It would be helpful to have some more packets of h265 frames

QuantumEntangledAndy:
The current implementation still will advance by 1byte scan forward in the event of an unknown data type. This only ever happens twice in my tests. Both at the start of the stream and both just after the InfoData packet. But other than that we have fully parsed stream kinds

QuantumEntangledAndy:
The reason for info packets failing is that their size field (bytes in 4..8) seem to be the size of the header not of the data to follow. Their is no data to follow. I have fixed this by setting the match in the get data size to always return 0.

I've been running the stream for a few minutes with no 1byte advancing the stream due to unknown or invalid packets.

yay

src/bc_protocol.rs
QuantumEntangledAndy: Done

src/bc_protocol.rs
QuantumEntangledAndy: Ok I have removed this and some other superfluous codes

QuantumEntangledAndy:
@thirtythreeforty If your free and have time for a review it would be appreciated I feel like it is ready to go unless you can spot anything odd.

QuantumEntangledAndy:
Also I have a working audio branch based off of this one yay

QuantumEntangledAndy:
Also have a video format autodetect based off of the audio branch

thirtythreeforty:
Wonderful, thanks. I definitely owe you a review; I am going to try to get to it later today.

QuantumEntangledAndy:
Thanks it would be nice to get this one done so I can put forward the audio and auto detect format branches.

src/bc_protocol/media_packet.rs
```diff
@@ -0,0 +1,381 @@
...
+            MediaDataKind::InfoData => 0, // The bytes in MediaData::bytes_to_size(&data[4..8]) seem to be the size of the header
+            MediaDataKind::Unknown | MediaDataKind::Invalid | MediaDataKind::Continue => data.len(),
+        }
+    }
+
+    pub fn data_size(&self) -> usize {
+        MediaData::data_size_from_raw(&self.data)
+    }
+
+    pub fn pad_size_from_raw(data: &[u8]) -> usize {
```
thirtythreeforty, line 114: Most of these functions do not need to be `pub`, because they are not used in any module other than this one.

src/bc_protocol/media_packet.rs
```diff
@@ -0,0 +1,381 @@
...
+}
+
+impl MediaData {
+    pub fn body(&self) -> &[u8] {
+        let lower_limit = self.header_size();
+        let upper_limit = self.data_size() + lower_limit;
+        let len = self.len();
+        if !self.complete() {
+            unreachable!(); // An incomplete packet should be discarded not read, panic if we try
+        }
```
thirtythreeforty, line 56: Since this function is `pub`, we should not assume that the client code will know to uphold this invariant. Is this `unreachable` because this module upholds the invariant? If so, this should be an `assert` (with a comment explaining why it's impossible).

src/bc_protocol/media_packet.rs
```diff
@@ -0,0 +1,381 @@
...
+        let header_size = MediaData::header_size_from_kind(kind);
+        if data.len() < header_size {
+            trace!("Magic failed header checks on size: {:x?}", &data);
+            return false;
+        }
+
+        match kind {
+            MediaDataKind::VideoDataIframe => {
+                let stream_type = &data[4..8];
+                if let Ok(stream_type_name) = str::from_utf8(stream_type) {
```
thirtythreeforty, line 203: It shouldn't be necessary to convert this to UTF8. You can simply match against the bytes, `b"H264"`, which is just sugar over the byte array.

src/bc_protocol/media_packet.rs
```diff
@@ -0,0 +1,381 @@
...
+                                &data[0..min(MAX_HEADER_SIZE, data.len())]
+                            );
+                            false
+                        }
+                    }
+                } else {
+                    false
+                }
+            }
+            MediaDataKind::VideoDataPframe => {
```
thirtythreeforty, line 219: The Iframe and Pframe cases are identical, collapse them.

src/bc_protocol/media_packet.rs
```diff
@@ -0,0 +1,381 @@
...
+        while self.binary_buffer.len() < MAX_HEADER_SIZE {
+            self.fill_binary_buffer(rx_timeout)?;
+        }
+
+        // Check the kind, if its invalid use pop a byte and try again
+        let mut magic =
+            MediaDataSubscriber::get_first_n_deque(&self.binary_buffer, MAX_HEADER_SIZE);
+        if INVALID_MEDIA_PACKETS.contains(&MediaData::kind_from_raw(&magic)) {
+            trace!("Advancing to next know packet header: {:x?}", &magic);
+        }
```
thirtythreeforty, line 317: I don't think you need this `if` block - it does the same check as the `while` beneath it.

src/bc_protocol/media_packet.rs
```diff
@@ -0,0 +1,381 @@
...
+
+    fn get_first_n_deque<T: std::clone::Clone>(deque: &VecDeque<T>, n: usize) -> Vec<T> {
+        // NOTE: I want to use make_contiguous
+        // This will make this func unneeded as we can use
+        // make_contiguous then as_slices.0
+        // We won't need the clone in this case either.
+        // This is an experimental feature.
+        // It is about to be moved to stable though
+        // As can be seen from this PR
+        // https://github.com/rust-lang/rust/pull/74559
```
thirtythreeforty, line 337: :+1:

src/bc_protocol/media_packet.rs
```diff
@@ -0,0 +1,381 @@
...
+        const MAGIC_PFRAME: &[u8] = &[0x30, 0x31, 0x64, 0x63];
+
+        let magic = &data[..4];
+        let kind = match magic {
+            MAGIC_VIDEO_INFO_V1 | MAGIC_VIDEO_INFO_V2 => MediaDataKind::InfoData,
+            MAGIC_AAC => MediaDataKind::AudioDataAac,
+            MAGIC_ADPCM => MediaDataKind::AudioDataAdpcm,
+            MAGIC_IFRAME => MediaDataKind::VideoDataIframe,
+            MAGIC_PFRAME => MediaDataKind::VideoDataPframe,
+            _ if data.len() == CHUNK_SIZE => MediaDataKind::Continue,
```
thirtythreeforty, line 164: This is the only thing I don't like about this current implementation. Should not assume that the sender is sending `CHUNK_SIZE` packets as a continuation - logically the media packet payloads simply form a byte stream and we should be able to read any incoming chunk size. (If this sounds a lot like `Read`, it's because it is.)

The better strategy is to keep adding whatever data is received to the outstanding packet.

That said, the API of this module is good - "get next media packet" makes it easy to change this after this PR is merged. So if you want me to look at it down the road, I'm okay with that, since it works with existing camera firmware.

QuantumEntangledAndy:
If you turn on trace level logging it will tell you if it's advancing frames due to unknown magic. Which is a sign of it dropping something.

src/bc_protocol/media_packet.rs
```diff
@@ -0,0 +1,381 @@
...
+        }
+        &self.data[lower_limit..upper_limit]
+    }
+
+    pub fn complete(&self) -> bool {
+        if self.len() < self.header_size() {
+            return false;
+        }
+        let full_data_len = self.len() - self.header_size();
+        full_data_len >= self.data_size() && full_data_len < self.data_size() + 8
```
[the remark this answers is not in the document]
QuantumEntangledAndy: I think I can take out some of those as there legacy from where we were learning the format. I'll go through them and see

src/bc_protocol/media_packet.rs
QuantumEntangledAndy: This if check is to print once only at the beginning of the loop not on every step of the loop.

src/bc_protocol/media_packet.rs
QuantumEntangledAndy: The actual implementation dosent depend on finding a continuation packet. It works like you say by pulling more data and adcanicng. so we could just drop this type.

QuantumEntangledAndy:
It should print the header of the unknow and if the first few bytes are around  the number thirty it might be a new header. Perhaps a B frame.

twistedddx:
If we could get about 15 seconds of BC media format from a B800/D800 it would be great to check the h265 format.
I know I only have h264 cameras and I think Andy also only has h264 cameras to play with.
The 15 seconds could be obtained by packet capture or a local Nvr export as "h264" format or via temp file official client creates prior to conversion to mp4.

src/bc_protocol/media_packet.rs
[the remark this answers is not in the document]
QuantumEntangledAndy: The 8 comes from 8 byte padding used in the stream format, I have added the `cost PAD_SIZE` to help with this

src/bc_protocol/media_packet.rs
QuantumEntangledAndy: I think I will take it out entirly as then we can removed the complete function too

src/bc_protocol/media_packet.rs
QuantumEntangledAndy: Done

src/bc_protocol/media_packet.rs
QuantumEntangledAndy: I've completely removed the continuation type as its all unneed and hidden by the get next media packet anyway

src/bc_protocol/media_packet.rs
QuantumEntangledAndy: Done

src/bc_protocol/media_packet.rs
```diff
@@ -0,0 +1,381 @@
...
+    }
+
+    pub fn expected_num_packet(&self) -> usize {
+        self.data_size() / CHUNK_SIZE + 1
+    }
```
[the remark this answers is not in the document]
QuantumEntangledAndy: Done

src/bc_protocol/media_packet.rs
QuantumEntangledAndy: Done

src/bc_protocol/media_packet.rs
QuantumEntangledAndy: Done

QuantumEntangledAndy:
I think I've got all your changes in please let me know what you think

QuantumEntangledAndy:
Also can we remove
```rust
#![allow(dead_code)]
#![allow(unused_variables)]
```
from lib.rs?

Then the linter will pick up unused functions are vars.

src/bc_protocol/media_packet.rs
QuantumEntangledAndy: I'd like to keep it in as its a good way to detect new magic and see when the code is skipping a packet. But if you still want to remove that sort of thing from the trace let me know and I'll axe it.

QuantumEntangledAndy:
Raw data may be useful especially if we have it from a few different camera, but for now just trace level should tell us when it finds a packet it dosen't recognise.

QuantumEntangledAndy:
I'd like to remove the full header check function too. It is there in case of the very small chance that 4 bytes are not enough to pick out the header in the video data as suggested by @hh1209. However I've never had this as neccessary and my back of the envelope calculation suggests that the odds of it happening randomly are 1 in 4,228,250,625 bytes. So once every 4GB.The advance header check is only really needed during the recovery phase when we are advancing by one byte at a time because we found an invalid packet. 99% of the time (as in all the time in my hour long test except during camera disconnection caused by my WiFi) we are not in the recovery phase so the chance of getting a magic packet out of place is even more unlikely then 1 in 4GB. 

Bottom line can we remove the advance header check and clean up code a bit more.

twistedddx:
> 
> 
> I'd like to remove the full header check function too. It is there in case of the very small chance that 4 bytes are not enough to pick out the header in the video data as suggested by @hh1209. However I've never had this as neccessary and my back of the envelope calculation suggests that the odds of it happening randomly are 1 in 4,228,250,625 bytes. So once every 4GB.The advance header check is only really needed during the recovery phase when we are advancing by one byte at a time because we found an invalid packet. 99% of the time (as in all the time in my hour long test except during camera disconnection caused by my WiFi) we are not in the recovery phase so the chance of getting a magic packet out of place is even more unlikely then 1 in 4GB.
> 
> Bottom line can we remove the advance header check and clean up code a bit more.

That would of been me. My suggestion is born out of my approach and code is absolutely terrible. I am a data cabler by trade and have written some code in my spare time.
My code was 100% based on brute force searching for magic bytes.
Although in my defense my code didn't get the luxury of the BC network protocol keeping the frames mostly split up in sane ways.

I agree that because your code is not dependent on brute force searching every byte for magic bytes this increased resilience to false positives may be over the top.

thirtythreeforty:
> I'd like to remove the full header check function too. It is there in case of the very small chance that 4 bytes are not enough to pick out the header in the video data as suggested by @hh1209. However I've never had this as neccessary and my back of the envelope calculation suggests that the odds of it happening randomly are 1 in 4,228,250,625 bytes. So once every 4GB.The advance header check is only really needed during the recovery phase when we are advancing by one byte at a time because we found an invalid packet. 99% of the time (as in all the time in my hour long test except during camera disconnection caused by my WiFi) we are not in the recovery phase so the chance of getting a magic packet out of place is even more unlikely then 1 in 4GB.
> 
> Bottom line can we remove the advance header check and clean up code a bit more.

Yes, that sounds okay to me. The packet stream seems reasonably well-formed, and your code copes with the cases where it's not.

thirtythreeforty:
Thank you for the changes removing the continuation kind. This looks great!

I would like to rescind my previous feedback about the grey frames and errors. I had managed to get my local Git client into a weird state where I thought I was up to date but I was actually running old code.  The tip of this branch is butter smooth for me in ffplay and VLC.  So yes, make any final changes you would like and I am good to merge this!

QuantumEntangledAndy:
Ah yes sorry, that's why I shouldn't tag from memory. Anyways while I agree this is a good idea for brute forcing every byte, we don't use that as a method and mostly use the size field given in the header.

Also revised estimate on the odds as I forget to take into account multiple types of magic header

| Magic | Byte 1 | Byte 2 | Byte 3 | Byte 4 |
|-------|-------|--------|--------|--------|
| InfoA  |   31    |    30     |     30    |     31    |
| InfoB  |   31    |    30     |     30    |     32   |
| AAC   |   30    |    35     |     77    |     62   |
| ADP   |   30    |    30     |     64    |     63   |
| Vid-I  |   30    |    31     |     63    |     63   |
|-------|-------|--------|--------|--------|
| Odds |  2 in 255 | 3 in 255 | 4 in 255 | 4 in 255|

Total Chance 1/255 * 3/255 * 4/255 * 4/255 = 1.13522E-08

Total 1 in 88,088,554 or 1 in 88MB

src/bc_protocol/media_packet.rs
thirtythreeforty: We can leave it but I reserve the right to refactor at some point if I change my mind ;)

src/bc_protocol/media_packet.rs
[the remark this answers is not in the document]
thirtythreeforty: Great, thanks

QuantumEntangledAndy:
Ok so that's it then all updated! We also remove the Invalid Media Kind too since the only thing that really checked that was this brute forcing

thirtythreeforty:
Once we realized it's an inner packet stream, have you actually seen any occurrences of truncated packets? We can take that probability to zero if it's parsing out well-formed packets :)

Also if there's a very very occasional truncated packet, the scan-for-magic-bytes behavior will smoothly pick the media packet stream back up after the corruption passes by.

QuantumEntangledAndy:
I also need a larger buffer even for my E1, maybe that needs a note on the readme

QuantumEntangledAndy:
I haven't actually seen a truncated packet even during my hour long test

QuantumEntangledAndy:
Well actually I did see it but only during a reconnect

thirtythreeforty:
One last change - can you add a `warn!()` to yell if we actually hit a truncated packet? If I never see any going forward, I am likely to remove that so there is no chance of accidentally truncating a packet that happens to start with magic bytes.

thirtythreeforty:
During a reconnect would probably be okay. In these kinds of edge cases the Unknown code would kick in. Or the media packet code could simply swallow the garbage until it hits a header.  Lots of mitigation behaviors are okay.

QuantumEntangledAndy:
We already have mitigation set up where we just drop the buffer coming from the camera until we hit magic again, so I think that's enough for such cases

thirtythreeforty:
Sweet. This has been quite the journey from "grep the extra junk out of the stream" to "rats there's another packet format" to "make a whole new parsing layer."

Thank you very much @twistedddx and @QuantumEntangledAndy

QuantumEntangledAndy:
Yeah it's been quite a journey. But I just want to say "Yay" merged!

thirtythreeforty:
I think audio will be a piece of cake now. If you have a branch you want to post, you can, or I can start on it

QuantumEntangledAndy:
Yes audio and auto detect are done. I am just rebasing onto master

PR #54, opened 2020-07-31
Automatically check/set the time on cameras, and bump timeout

### LINKED ISSUE #14: Timeout loop after update/reboot/disconnect/crash
When a camera is suddenly disconnected from the network by unplugging a cable, or Neolink crashes or is killed, sometimes the disconnected camera(s) will no longer stream with Neolink's stock timeout threshold once the camera is reconnected. Neolink will try to connect, will successfully reconnect, then will immediately time out, then reconnect again successfully, then time out again, and repeat the cycle indefinitely. This disconnect/timeout problem only happens sometimes, and very rarely. Most of the time, I can pull the plug on any of the cameras, reconnect, and everything works fine.

I can manually set the timeout threshold to the timeout-ing camera to 5 seconds in the Neolink config file, then it streams perfectly.

I also noticed that (on the rare occasion) after the camera is disconnected and begins to time out, the time and date overlay will disappear from the top of the feed. The date is reset to the year 2000 in the settings, and it won't display it on the feed until I sync the clock with the computer using the Reolink app.

Once I sync the system time with the camera, the date/time overlay will return to the top of the feed, and I can turn the timeout back down to one second in the Neolink config file for that camera and it streams perfectly, without any timeouts.

I'm not sure what's causing the camera's date and time settings to be destroyed, but I think that is what is causing the timeout issue. A workaround for this might be to figure out how the time is set on the camera via the Baichuan protocol, then have Neolink sync the system time with the camera on every successful reconnect.

thirtythreeforty:
Collectively, this should close #14.

### DIFF
--- a/src/bc_protocol/time.rs
+++ b/src/bc_protocol/time.rs
@@ -1,4 +1,4 @@
-use super::{BcCamera, Error, Result};
+use super::{BcCamera, Error, Result, RX_TIMEOUT};
 use crate::bc::{model::*, xml::*};
 use time::{date, Date, OffsetDateTime, PrimitiveDateTime, Time, UtcOffset};
 
@@ -20,7 +20,7 @@ impl BcCamera {
         };
 
         sub_get_general.send(get)?;
-        let msg = sub_get_general.rx.recv_timeout(self.rx_timeout)?;
+        let msg = sub_get_general.rx.recv_timeout(RX_TIMEOUT)?;
 
         if let BcBody::ModernMsg(ModernMsg {
             xml:
@@ -105,7 +105,7 @@ impl BcCamera {
         );
 
         sub_set_general.send(set)?;
-        let msg = sub_set_general.rx.recv_timeout(self.rx_timeout)?;
+        let msg = sub_set_general.rx.recv_timeout(RX_TIMEOUT)?;
 
         Ok(())
     }
--- a/src/bc_protocol/time.rs
+++ b/src/bc_protocol/time.rs
@@ -1,0 +1,132 @@
+use super::{BcCamera, Error, Result};
+use crate::bc::{model::*, xml::*};
+use time::{date, Date, OffsetDateTime, PrimitiveDateTime, Time, UtcOffset};
+
+impl BcCamera {
+    pub fn get_time(&self) -> Result<Option<OffsetDateTime>> {
+        let connection = self
+            .connection
+            .as_ref()
+            .expect("Must be connected to get time");
+        let sub_get_general = connection.subscribe(MSG_ID_GET_GENERAL)?;
+        let get = Bc {
+            meta: BcMeta {
+                msg_id: MSG_ID_GET_GENERAL,
+                client_idx: 0,
+                encrypted: true,
+                class: 0x6414,
+            },
+            body: BcBody::ModernMsg(ModernMsg::default()),
+        };
+
+        sub_get_general.send(get)?;
+        let msg = sub_get_general.rx.recv_timeout(self.rx_timeout)?;
+
+        if let BcBody::ModernMsg(ModernMsg {
+            xml:
+                Some(BcXml {
+                    system_general:
+                        Some(SystemGeneral {
+                            time_zone: Some(time_zone),
+                            year: Some(year),
+                            month: Some(month),
+                            day: Some(day),
+                            hour: Some(hour),
+                            minute: Some(minute),
+                            second: Some(second),
+                            ..
+                        }),
+                    ..
+                }),
+            ..
+        }) = msg.body
+        {
+            let datetime = match try_build_timestamp(
+                time_zone, year, month, day, hour, minute, second
+            ) {
+                Ok(dt) => dt,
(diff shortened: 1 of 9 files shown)


thirtythreeforty:
@hh1209 can you download the prebuilt for this PR and see if this fixes the bootlooping camera problem? Stress test as best you can.

thirtythreeforty:
One thing I noticed is if the cams have internet access, they'll auto set their times via NTP very quickly after you try to change it with the set_time API call.

thirtythreeforty:
@QuantumEntangledAndy I'd appreciate a code review since you're rapidly getting familiar with the project.

src/bc_protocol/time.rs
```diff
@@ -0,0 +1,132 @@
...
+                    reply: msg,
+                    why: "Could not parse date",
+                })
+            };
+
+            // This code was written in 2020; I'm trying to catch all the possible epochs that
+            // cameras might reset themselves to. My B800 resets to Jan 1, 1999, but I can't
+            // guarantee that Reolink won't pick some newer date.  Therefore, last year ought
+            // to be new enough, yet still distant enough that it won't interfere with anything
+            const BOUNDARY: Date = date!(2019-01-01);
```
QuantumEntangledAndy, line 58: How far out of sync does reolink cameras let us be before we get dissconnects? It may be sufficient to just have this as NOW-10mins I assume that anyone running a camera would like the correct time stamps on there videos.

src/config.rs
```diff
@@ -51,6 +51,7 @@ pub struct CameraConfig {
     pub username: String,
     pub password: Option<String>,
 
+    // no longer used, but still here so we can warn users:
     pub timeout: Option<Duration>,
```
QuantumEntangledAndy, line 55: Rather than a warning what about a validation error and it exits? Or am I being too mean..

src/bc_protocol/time.rs
```diff
@@ -0,0 +1,132 @@
...
+            }
+        } else {
+            Err(Error::UnintelligibleReply {
+                reply: msg,
+                why: "Reply did not contain SystemGeneral with all time fields filled out",
+            })
+        }
+    }
+
+    pub fn set_time(&self, timestamp: OffsetDateTime) -> Result<()> {
```
QuantumEntangledAndy, line 75: Will setting time fail if we don't have admin priv?

QuantumEntangledAndy:
p.s. Are you sure I should be code reviewing I don't feel like my rust foo is of any great level

src/config.rs
thirtythreeforty: Since it's otherwise harmless to have there, I'd say I prefer a graceful migration. Now if it was an "action required" behavior change, I would agree.

src/bc_protocol/time.rs
thirtythreeforty: Oh that's a good question. I have not tried that.

src/bc_protocol/time.rs
thirtythreeforty: Hm that's a good point. Maybe I should take that approach and let someone yell if they don't like it.

thirtythreeforty:
This sets the time twice if you have `mainStream` and `subStream` enabled. One of the streams needs to be instructed to be the "management" stream that does all this housekeeping.

hh1209:
> One of the streams needs to be instructed to be the "management" stream that does all this housekeeping.

Just a thought, what if someone is only using one stream, like on 400 series cams? i.e. not the "management" stream

thirtythreeforty:
> Just a thought, what if someone is only using one stream, like on 400 series cams? i.e. not the "management" stream

Neolink would ensure one of the connections is always a management connection.

src/bc_protocol/time.rs
thirtythreeforty: Well, it doesn't really matter what's going to happen here if we don't have permission. We do recheck that the action "took" and print an error if not. So I think that's okay.

src/config.rs
QuantumEntangledAndy: Fair enough, since its undocumented anyways I guess we can just do it that with a warning

src/bc_protocol/time.rs
QuantumEntangledAndy: I see, then am wondering if it would help the user for that message to give a potential reason for not being able setting it. Else we may have issue reports asking for help with their error.

src/bc_protocol/time.rs
thirtythreeforty: So I am actually going to leave this particular snippet the way it is, because it's detecting time-not-set-at-all, not time-wrong.  It should be the job of the caller of this function to detect time-wrong.

thirtythreeforty: I don't dispute that fixing time-wrong would be nice.

src/bc_protocol/time.rs
thirtythreeforty: Good point. Done.

PR #59, opened 2020-08-02
Add macos build and more architectures

QuantumEntangledAndy:
This adds a macos build. I also tried to add stretch but it seems that our

```toml
gstreamer-rtsp-server = { version = "0.15", features = ["v1_12", "v1_14"]}
```

Bumps up the required version from the default "v_08" (according to their readme) to "v_14" and therefore won't work for stretchs "v_10" of gstreamer. We need `"v_14"` for the auth code so unless we have some sort of optional without auth there won't be a stretch build.

I did however in the process make armhf use matrix which may be useful in the future so I kept it in this PR


thirtythreeforty:
For the armhf matrix, stretch would be possible if authentication were behind a feature flag (then just turn it off when building for <1.14). I think I don't want that complexity in the code though - 1.14 has been around for a while, and supporting Debian's oldstable means you'll be foregoing a lot of years of features.

QuantumEntangledAndy:
That's what I was feeling like too stretch is using quite an old gst and I'm not sure we'd want to support that.

I'm also wondering if their be demand for a brewfile (macos package) then installing neolink would be as simple as

```bash
brew install neolink
```

thirtythreeforty:
I have no problem with this - just haven't gotten around to saying so... Would you pin the Gstreamer dependencies following the same lessons learned as on Windows (#62)?

QuantumEntangledAndy:
There's another issue with this that I ran into when working on the Mac package. If I use brew for installing gstreamer then the binary will expect to find the libraries in /use/local/. If I use gstreamers own PKG installer then the binary will expect to find the libraries in /Frameworks/Gstreamer.

On mac depency paths are baked in as absolute paths at build time. We can change it to something relative like excutable_path and rpath [here](https://medium.com/@donblas/fun-with-rpath-otool-and-install-name-tool-e3e41ae86172) but not sure that would work. 

Which dependency install method would work best?

Should we:

1. Compile against the gstreamer pkg

2. Compile against the brew package

---

As a compramisr we could provide two ways to install.
1. From this repo CI as normal which will be the gstreamer pkg version
2. From homebrew using
```
brew install neolink
```
Which will be the homebrew linked version. 

For this to work as I described we would have to

1. Change this PR to use the pkg
2. Submit a brewfile to homebrewcore with our program
3. Add a build step that will automatically send an update request to homebrewcore when a new tag is created. 


---

We could also forgo having our own binary and just submit to homebrew core.

QuantumEntangledAndy:
Given what is happening in #66 and #76 I have added a few more build here namely

- Mac os uses the gstreamer from the msi, since I feel most people will use that for installing gstreamer and once the pkg is merged they would use that.

- I moved the armhf build into the build.yaml so that users only need to go to one CI work flow for all the artifacts.

- I added the other archtectures that I worked up for #76 I am not sure they are needed at all and am quite happy to revert the commit and then if reverted at least we we will have it in the git history if we every need it.

QuantumEntangledAndy:
Could we get this merged? We have a few mac users now and I have had a message or two about `"where is the raspberry pi build?"`, which this PR addresses too by putting it in the same workflow as CI

.github/workflows/build.yaml
```diff
@@ -47,3 +61,143 @@ jobs:
...
+          CARGO_TARGET_I686_UNKNOWN_LINUX_GNU_LINKER: i686-linux-gnu-gcc
+          CC_i686_unknown_linux_gnu: i686-linux-gnu-gcc
+          CXX_i686_unknown_linux_gnu: i686-linux-gnu-g++
+          PKG_CONFIG_ALLOW_CROSS: 1
+          PKG_CONFIG_PATH: /usr/lib/i386-linux-gnu/pkgconfig:/usr/share/pkgconfig/
+          PKG_CONFIG_LIBDIR: /usr/lib/i386-linux-gnu/pkgconfig:/usr/share/pkgconfig/
+      - uses: actions/upload-artifact@v2
+        with:
+          name: release-i386-buster
+          path: "target/i686-unknown-linux-gnu/release/neolink*"
```
thirtythreeforty, line 203: Is it possible to parameterize these into a single template (maybe "`ubuntu-cross-build`") that gets instantiated multiple times?

thirtythreeforty (changes requested):
Thanks for this. Yes, I'm glad to add support. One question:

.github/workflows/build.yaml
QuantumEntangledAndy: I'm not sure about a template maybe a custom action? Not sure. I could do it using shell scripting instead. Maybe have a script that would take as input the arch and which then calls `cargo`. I think there are a few places we could reuse it in the packaging PR too.

I could also doit using some output variables generated in a prior action. But we are just moving the definitions elsewhere and I'm not sure it will help readability.

If I get the time I'll have a look into a generic build script that can be used with this and my packaging PR

thirtythreeforty: I'm talking about using [contexts and a matrix](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions). If you don't mind, I'd like to take a shot at refactoring this. I would like to get more proficient at GH Actions.

thirtythreeforty: And, would you mind if I squash and rebase some of these commits?

QuantumEntangledAndy: Squash and refractor away :).

thekiefs:
> Given what is happening in #66 and #76 I have added a few more build and changed macos gstreamer dependency here namely
> 
>     * Mac os uses the gstreamer from the msi, since I feel most people will use that for installing gstreamer and once the pkg is merged they would use that which installs gstreamer from the msi too, this also pins it to a specific version.
> 
>     * I moved the armhf build into the build.yaml so that users only need to go to one CI work flow for all the artifacts.
> 
>     * I added the other archtectures that I worked up for [Multi Architeture docker #76](https://github.com/thirtythreeforty/neolink/pull/76) I am not sure they are needed at all and am quite happy to revert the commit and then if reverted at least we we will have it in the git history if we every need it.

I'm on MacOS

`brew install gstreamer gst-plugins-good gstreamer1.0-plugins-good gst-plugins-base`
`sudo cp neolink /usr/local/bin/ `                                            
`sudo chmod +x /usr/local/bin/neolink`

However get the error

```
neolink rtsp --config=config.yaml                                                                                      
dyld[21373]: Library not loaded: /Library/Frameworks/GStreamer.framework/Versions/1.0/lib/libgstrtspserver-1.0.0.dylib
  Referenced from: /usr/local/bin/neolink
  Reason: tried: '/Library/Frameworks/GStreamer.framework/Versions/1.0/lib/libgstrtspserver-1.0.0.dylib' (no such file), '/usr/local/lib/libgstrtspserver-1.0.0.dylib' (no such file), '/usr/lib/libgstrtspserver-1.0.0.dylib' (no such file)
```

QuantumEntangledAndy:
MacOS stores the absolute path to dependencies upon build. Brew installs those dependencies in `/usr/local` while gstreamers dpkg installer does it to `/Library/frameworks/`. Because these are different paths they are incompatible. 

You need to install gstreamer using the dpgk installer on their website. Or build neolink from source.

PR #65, opened 2020-08-09
Add queues to the gstreamer launch

### LINKED ISSUE #62: Queue needed in gst
**Describe the bug**
Gst reports that there is not enough buffers when connecting long distance over UDP (ssh tunnel on the bus over 4G fyi)

**To Reproduce**
Steps to reproduce the behavior. Example:
1. Set GST_DEBUG to 3
2. Launch neolink over a latent connection
3. Connect to neolink over UDP, (TCP seems fine)

**Expected behavior**
Gst debug logs not to report a warning like this

```cpp
WARN                basesink gstbasesink.c:1218:gboolean gst_base_sink_query_latency(GstBaseSink *, gboolean *, gboolean *, GstClockTime *, GstClockTime *):<multiudpsink0> warning: Not enough buffering available for  the processing deadline of 0:00:00.020000000, add enough queues to buffer  0:00:00.020000000 additional data. Shortening processing latency to 0:00:00.000000000.
```

We should probably just add a queue to the gst launch command. I will work up a PR when I get time and this is to keep track of the issue until then

QuantumEntangledAndy:
This adds queues to the gstreamer launch command

fixes #62 

This was prompted by gstreamer issuing warning due to missing queues when using UDP over a slow connection.

### DIFF
--- a/src/gst.rs
+++ b/src/gst.rs
@@ -53,8 +53,8 @@ impl RtspServer {
             .expect("The server should have mountpoints");
 
         let launch_str = match stream_format {
-            StreamFormat::H264 => "! h264parse ! rtph264pay name=pay0",
-            StreamFormat::H265 => "! h265parse ! rtph265pay name=pay0",
+            StreamFormat::H264 => "! queue ! h264parse ! rtph264pay name=pay0",
+            StreamFormat::H265 => "! queue ! h265parse ! rtph265pay name=pay0",
             StreamFormat::Custom(ref custom_format) => custom_format,
         };
 


thirtythreeforty:
This is fine but needs a max buffer size in the launch command, or the queue will happily grow to gigabytes in case there's a problem.

QuantumEntangledAndy:
really? The [documentation](https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer-plugins/html/gstreamer-plugins-queue.html) suggests a default max size of 10485760 bytes. Which is about 10MB. Am I reading it wrong?

QuantumEntangledAndy:
After it reaches the max buffer size it suspends the sink thread.

However our sink thread is an appsrc, which have its own buffer that fills..

Have you tested this. I did notice awhile ago you had a queues branch.

QuantumEntangledAndy:
Since the queue will be merged as part of audio PR I will close this one

Issue #63, opened 2020-08-05
Use correct timestamps


QuantumEntangledAndy:
**Is your feature request related to a problem? Please describe.**
Currently we use the clock for the time stamps. This is not ideal, it would be get the timestamp from the header of media packets (bytes 16-20 of the I and P frames).

**Describe the solution you'd like**
Use the `set_dts` and `set_pts` in the `gstreamer::Buffer` object ([described here](https://github.com/sdroege/gstreamer-rs/blob/01eaf76b31701f23c6dcac3f84cb48e93864a639/gstreamer/src/buffer.rs)) to set the time stamps.

PR #68, opened 2020-08-12
Audio and Format auto detect

### LINKED ISSUE #4: Stream audio in same RTSP feed
The cameras optionally support audio - this is disabled by default. However it should be possible to mux this into the same RTSP feed. More reverse engineering is needed to learn the Baichuan format for audio.

### LINKED ISSUE #36: Autodetect camera type and stream type
Users should not have to know or care what type of camera they're connecting to, or what kind of video/audio codec it's using.  Autodetect these.

### LINKED ISSUE #62: Queue needed in gst

QuantumEntangledAndy:
This PR adds...... Audio and format auto detection :).

To change the factory launch after initialisation for to the auto detected format I needed to keep the factory object around, so I created the `GstOutputs` struct which controls the outputs into gstreamer.

This `GstOutputs` structure also contains the appsrcs yes plural, we have one for audio and one for video now so I needed to pass them around as a set.

I couldn't see a ADPCM RTP format in gstreamer but I could find linear pcm ones. I am hoping they will suffice. Using these for reference: [rawaudioparse format=pcm](https://gstreamer.freedesktop.org/documentation/rawparse/rawaudioparse.html?gi-language=c#page-description) and [rtppcmapay](https://gstreamer.freedesktop.org/documentation/rtp/rtppcmapay.html?gi-language=c#rtppcmapay). My E1 is AAC so I cannot test, but perhaps @twistedddx could test for us.

I have also added queues which means my other PR #65 is superseded by this one.

Closes #4
Closes #36
Closes #62

Closes #65


QuantumEntangledAndy:
Oh and I also applied clippy styling. Let me know if that too many changes to keep track of and I'll drop that commit.

QuantumEntangledAndy:
Also I have a question about how rust handles destruction of objects and the gst factory. 


In `add_stream` we used to create the factory object as a local variable. Since this was owned by a local variable was it not destroyed at the end of the function.

If it was destroyed at the end of the function, then how were we able to use it in the callback where we fish out the appsrc from the media?

I am wondering if perhaps it was destroyed but gst being less safe Cpp code kept using it. And that this is causing our segfault....

twistedddx:
AAC Audio works for:
RLC-410 4MP v1 & v2
RLC-410 5MP
RLC-520 5MP

Swann NHD-818 4MP which uses ADPCM has issues:
C:\gstreamer\1.0\x86_64\bin>neolink.exe --config neolink_config.toml
[2020-08-14T14:25:13Z INFO  neolink] Neolink v0.2.0-158-g6714847 release
[2020-08-14T14:25:13Z INFO  neolink] test: Connecting to camera at 192.168.93.2:53719
[2020-08-14T14:25:13Z INFO  neolink] test: Connected and logged in
[2020-08-14T14:25:13Z INFO  neolink] test: Camera time is already set: 2020-08-14 23:54:36 +9:30
[2020-08-14T14:25:13Z INFO  neolink] test: Starting video stream mainStream
[2020-08-14T14:25:13Z WARN  neolink::bc_protocol::media_packet] Possibly truncated packet or unknown magic in stream
[2020-08-14T14:25:13Z WARN  neolink::bc_protocol::media_packet] Possibly truncated packet or unknown magic in stream
[2020-08-14T14:25:13Z WARN  neolink::bc_protocol::media_packet] Possibly truncated packet or unknown magic in stream
[2020-08-14T14:25:14Z WARN  neolink::bc_protocol::media_packet] Possibly truncated packet or unknown magic in stream

unknown magic warning spams until Neolink is stopped.

The NHD-818 stream appears to include 00dc and 01wb as expected:
00dc
01wb

note all these cameras are being tested via Reolink NVR.

QuantumEntangledAndy:
Could you send the wireshark dump (only export msgid type 3 using a filter then the option during export) (also record a blank wall in a quite place as it contains the I and P frames). I want to see why you get so many interrupted packets. It implies we are not parsing the media packets right.

QuantumEntangledAndy:
What gets me with the audio packets is that the header is `<magic><size1><size2>` where size 1 and 2 are the same. PCM is sampled analogue (representing a wave sampled at intervals) I am wondering if this is maybe to do two channel stereo, it needs two waves and therefore two streams.

QuantumEntangledAndy:
p.s. this is the filter `baichuan.msg_id == 3`

twistedddx:
Andy I have emailed a link to a packet capture.
From what I see I think these Swann cameras may not be separating frames into separate msgid 3 payloads.
It appears like a continuous stream of the BC media data.

The audio size thing bothered me also.
Theories I came up with is:
1)Some code bug when writing 2 bytes of data for 4 bytes. This product doesn't appear full of perfect code.
2)The format allowed for actual audio bytes vs data size sent. Maybe the inventor thought there could be a case of 120 bytes of audio but they know they sent 128 bytes of data.

QuantumEntangledAndy:
Thanks I'll look at when I get home. Hopefully can figure this out. If you have tine could you also send the log file from neolink when GST_DEBUG=3 and RUST_LOG=trace that will tell me more about what's happening inside neolink.

QuantumEntangledAndy:
In bash (Linux or osx or Windows Linux subsystem) you can do this:

```bash
GST_DEBUG=3 RUST_LOG=trace ./neolink --config ... 2>&1 > debug.log
```

twistedddx:
https://pastebin.com/e8vd6H7j

The first 3 problem packets are the second half of the BC header for 01wb
00, 01, 7a, 00

QuantumEntangledAndy:
@twistedddx  I think you sent the wrong, pcapng the one you sent is missing all of the bc packets

Screenshot 2020-08-15 at 18 03 50

QuantumEntangledAndy:
@twistedddx could you also test my latest version

twistedddx:
> 
> 
> @twistedddx I think you sent the wrong, pcapng the one you sent is missing all of the bc packets
> 

The packet capture does not capture the initial connection to the camera but definitely includes the ongoing video/audio stream of BC media.
Do note the port is not the 9000 but instead 53719.

pcap

twistedddx:
New builds log:
https://pastebin.com/hrVZMXC1

twistedddx:
This is what neolink logs when vlc or mpc attempt to connect while the unknown magics are occurring.
https://pastebin.com/MpAMNUjC

QuantumEntangledAndy:
So from what I can see the ADPCM header size is 8 bytes not the 16 suggested in [the website](https://www.wasteofcash.com/BCConvert/BC_fileformat.txt)

QuantumEntangledAndy:
This newer version now uses just 8bytes

QuantumEntangledAndy:
@twistedddx  another test if you could, not sure what your time zone is so no worries if its an odd time for you whenever you have the time is fine

QuantumEntangledAndy:
Oh right and I force pushed. If your pulling from git it may complain. Might be easier to delete your local copy of my branch and then fetch it again.

```
git checkout master
git -D detect
git checkout -t origin detect
```

twistedddx:
Latest code with 8 byte 01wb header size still does exactly the same with unknown magic's.
I am +9:30, we still got plenty of time :)

In my example
30 31 77 62 F8 00 F8 00 00 01 7A 00 58 FB 31 00


00 01 7A 00 58 FB 31 00
I believe is not actual ADPCM data from my testing.
My opinion is it is meta data about the ADPCM data.
In my testing with ffmpeg that data created noise and increases the audio length beyond the video stream length.
The first 4 bytes there is in every single frame. the next 4 bytes are described by a Russian dev with a QCam BC device as being ADPCM step_index and current. I believe these are required to initialize the ADPCM decoder correctly. Without that data the audio output is not going to be as accurate as possible.

Have you forced a refresh of my filformat txt as I have updated it and at least my browser cached an older version.
F8 is 248 should be data size
7A is 122 which may be samples?
If 122 is samples it does allow for 122+122 as L+R which could fit in the 248 data size.
But in my experience with ffmpeg taking all the 248 bytes data as mono audio the audio length was correct.

Packet 120 in the capture is nice because it includes 2 full 01wb frames in the single packet.

I have tried to pretty the data slightly here:
https://pastebin.com/tgEufPNL
To the right I have added byte count columns.
First column is total 01wb frame byte count and second column is my count assuming a 16 byte header.
"|" Shows start of audio going by my version of the spec.
* shows where 122 would end.

QuantumEntangledAndy:
I see. It's just that: based on the size field (it was f8 I think) and that  there were 4+f8 bytes from the start to the next header start it suggests the header is 4 bytes. So I need to have a header size of 4 bytes if I want to calculate the correct step to the next header.

I will keep digging. I'm UCT+7 but heading to bed soon as that's what baby wants.

QuantumEntangledAndy:
Are the unknow magic codes immediately after the message about `Possibly truncated packet or unknown magic in stream` the same or random?

twistedddx:
Random

twistedddx:
Oh f0 de bc 0a 03 00 00 00 is not part of the payload.
I don't know why I doubt you guys ever!

The below shows a 03wb that is A8 long (168 bytes)
https://hsto.org/webt/fx/vf/4r/fxvf4rhqkxhhbkyy84gqiwuajoq.png

164 bytes is the ADPCM payload in purple.
The 4 more bytes would be 00 01 52 00 which as my 00 01 7A 00 equal.
52 is 82, which again is data size = samples? * 2 + 4.

QuantumEntangledAndy:
That's interesting. Because when we started at 16 bytes we were getting the same magic during the recovery phase. Now that we've gone to 8 I would expect it to be still not random. Unless. .... 

Are you compiling from source? If so can you please confirm that you reverted my forced push changes. This is in the function data_size_from_raw on line 71 in src/bc_protocol/media_packet.rs that the ADPCM field is something like this

```rust
MediaDataKind::AudioDataAdpcm => MediaData::bytes_to_size(&data[4..6]),
```

twistedddx:
Sorry I missed you said delete your branch and fetch again.
After doing that there is no more bad magic's.

But VLC still doesn't connect:
https://pastebin.com/LimiAbK2

QuantumEntangledAndy:
Oh cool. That's actually good news. It means we are parsing correctly. We just need to work out the correct launch str for gstreamer.

QuantumEntangledAndy:
I need to go to bed but you can play with the launch str (it's near the top of gst.rs and looks something like `rawparse format=pcm....`) to get it to work. (playing with the launch str for h264 is how I started my journey into rust and this project). I recommend having RUST_LOG set to default (warn I think) and just run neolink with GST_DEBUG=3 it will tell you which part of the gst launch string is not behaving.

QuantumEntangledAndy:
The rawparse is supposed to be for linear pcm I think which may not be the right choice. We will probably have to go through the gstreamer docs looking for a better parser.

twistedddx:
using fakesink for the audio does allow vlc to play the video although some invalid nal warnings can occur.

I assume something like this isn't far from the right answer:
Some(StreamFormat::ADPCM) => "! queue ! avdec_adpcm_ima_oki ! alawenc ! rtppcmapay name=pay1",

But that just gives me this on vlc connect
https://pastebin.com/uNjfyGm2

QuantumEntangledAndy:
Yes I also get the nal warning for my h264 but only the first few frames we used to get the constantly which is what prompted me to filter out the audio in my previous PR.

The launch strs you are using seem reasonable. If we want to convert (which is what you are doing) then I was thinking this:

```
! adpcmdec ! faac ! rtppcmapay name=pay1
```

It would be better to avoid conversion and just parse the stream on if we can.

QuantumEntangledAndy:
If you dump the raw stream to file with this

```
filesink location=dump.adpcm
```

and post it i can help find the gst str too using it

QuantumEntangledAndy:
This is what your one was doing
```
! queue ! avdec_adpcm_ima_oki ! alawenc ! rtppcmapay name=pay1
queue -> decode from binary raw to binary adpcm -> encode from binary to alaw -> send to rtp
```

What we want is a parser that dose this
```
queue -> parse binary to adpcm -> rtp
```

QuantumEntangledAndy:
We may also be able to use capsetter

```
! capssetter join=false replace=true caps="audio/x-adpcm"
```

instead of parsing but we need to know things like the sample rate too I think

QuantumEntangledAndy:
Ah I just found this

```
! wavparse
```

Which according to the [docs](https://gstreamer.freedesktop.org/documentation/wavparse/index.html?gi-language=c) can do 

` audio/x-wav: -> audio/x-adpcm:` which may handle our parsing to adpcm correctly hmmmm

QuantumEntangledAndy:
Hmm maybe we could try

```
! queue ! wavparse ! alawenc ! rtppcmapay name=pay1
```

Would be nice if we had a adpcm rtp rather then an alaw one though

QuantumEntangledAndy:
Aha found it!

```
rtpg726pay
```

Is the rtp payloader for adpcm

QuantumEntangledAndy:
So looking it adpcm it comes in many layouts...

audio/x-adpcm
- swf
- yamaha
- ea-xas
- xa
- dvi
- microsoft
- g726
- g721,
- g722
- g723_3
- g723_5

We probably need to figure out which one this one is....

QuantumEntangledAndy:
@twistedddx you gave this ffmpeg in #4

```
ffmpeg.exe -f g729 -c:a adpcm_ima_oki -i _BC_raw.audio -y _BC_raw.aac
```

are you suggesting that the input is in g729 format?

QuantumEntangledAndy:
It would be interesting to see the output of

```
ffprobe _BC_raw.audio
```

To see what ffmpeg autodetects the format as

twistedddx:
[dump.zip](https://github.com/thirtythreeforty/neolink/files/5079996/dump.zip)

Here is the result from "! filesink location=dump.adpcm"

ffprobe.exe dump.adpcm
ffprobe version git-2019-10-19-31aafda Copyright (c) 2007-2019 the FFmpeg developers
  built with gcc 9.2.1 (GCC) 20191010
  configuration: --enable-gpl --enable-version3 --enable-sdl2 --enable-fontconfig --enable-gnutls --enable-iconv --enable-libass --enable-libdav1d --enable-libbluray --enable-libfreetype --enable-libmp3lame --enable-libopencore-amrnb --enable-libopencore-amrwb --enable-libopenjpeg --enable-libopus --enable-libshine --enable-libsnappy --enable-libsoxr --enable-libtheora --enable-libtwolame --enable-libvpx --enable-libwavpack --enable-libwebp --enable-libx264 --enable-libx265 --enable-libxml2 --enable-libzimg --enable-lzma --enable-zlib --enable-gmp --enable-libvidstab --enable-libvorbis --enable-libvo-amrwbenc --enable-libmysofa --enable-libspeex --enable-libxvid --enable-libaom --enable-libmfx --enable-ffnvcodec --enable-cuvid --enable-d3d11va --enable-nvenc --enable-nvdec --enable-dxva2 --enable-avisynth --enable-libopenmpt --enable-amf
  libavutil      56. 35.101 / 56. 35.101
  libavcodec     58. 59.102 / 58. 59.102
  libavformat    58. 33.100 / 58. 33.100
  libavdevice    58.  9.100 / 58.  9.100
  libavfilter     7. 63.100 /  7. 63.100
  libswscale      5.  6.100 /  5.  6.100
  libswresample   3.  6.100 /  3.  6.100
  libpostproc    55.  6.100 / 55.  6.100
dump.adpcm: Invalid data found when processing input

ffplay.exe -f g729
unrecognizable audio

ffplay.exe -f g726
unrecognizable audio

ffmpeg.exe -f g729 -c:a adpcm_ima_oki -i dump.adpcm -y dump.aac
dump.aac becomes recognizable but clearly has lots of unknown noise added.

QuantumEntangledAndy:
Thanks I'll try to figure out the layout and work from there.

twistedddx:
[bcconvert.zip](https://github.com/thirtythreeforty/neolink/files/5080006/bcconvert.zip)

And here is the data my tool strips out of the file the official client creates at C:\RepTemp\.bctmp\download
The noise in the dump.adpcm is likely the first 8 bytes per frame.

twistedddx:
Note the dump.adpcm and download_BC_raw.audio do include an overlap of the same audio these 2 files start and end at non identical times.

twistedddx:
I am unsure what format the audio is exactly.
I have seen people say many CCTV are using g726 but not about BC particularly.
The hisilicon SDK includes the headers for hi_VoiceEngine which from memory mentioned g726 and adpcm.

I came up with ffmpeg.exe -f g729 -c:a adpcm_ima_oki -i dump.adpcm -y dump.aac via brute force of similar looking codecs of what I thought I might have. Basically the exact path we are on with gstreamer at the moment.

-f g729 may just be a fluke prep of ffmpeg in to what is needed.
Eg endianness, size and type of samples, samplerate etc.

I think conversion to a common format like aac would be wise as player support for adpcm is going to be poor.

QuantumEntangledAndy:
Yes its probably best to convert but we still need to find the right parser to set the gst caps correctly before conversion.

btw ffplay excepts all the same **input** arguments as ffmpeg so you can check out the wave with

```
ffplay -f g729 dump.adpcm
```

QuantumEntangledAndy:
Do you think:

```
ffplay -f g729 dump.adpcm
```

and 

```
ffplay -f g729 download_BC_raw.audio
```

sound pretty much the same?

QuantumEntangledAndy:
g726 also plays but is just white noise

twistedddx:
I couldn't work out how to tell ffplay to use adpcm_ima_oki

This works with clean audio:
ffmpeg.exe -f g729 -c:a adpcm_ima_oki -i download_BC_raw.audio -y download_BC_raw.aac

QuantumEntangledAndy:
I believe -c:a adpcm_ima_ok is an output option

QuantumEntangledAndy:
I recind my previous statement

```
ffplay -f g729 -acodec adpcm_ima_oki download_BC_raw.audio
```

QuantumEntangledAndy:
These all seems to work

```
ffplay -f g722 -acodec adpcm_ima_oki download_BC_raw.audio
ffplay -f g723_1 -acodec adpcm_ima_oki download_BC_raw.audio
ffplay -f g726 -acodec adpcm_ima_oki download_BC_raw.audio
ffplay -f g726le -acodec adpcm_ima_oki download_BC_raw.audio
ffplay -f g729 -acodec adpcm_ima_oki download_BC_raw.audio
```

Sigh....

QuantumEntangledAndy:
Now do they work because we have the right one or because we are relying on adpcms error recovery methods

twistedddx:
ffplay -f s16le -ar 8000 -acodec adpcm_ima_oki download_BC_raw.audio

I think the g729 just told ffmpeg the input is s16 and sample rate is 8000

QuantumEntangledAndy:
Hmm pretty much any combination of pcm -f option will work

```
ffmpeg -formats | grep -i pcm
```

- u8
- s8
- s24le
- s16le
- s32le
- f32le
- etc

QuantumEntangledAndy:
So having a go at gst and this is the only one I can get to link so far

```
gst-launch-1.0 filesrc location=dump.adpcm  ! rawaudioparse format=pcm pcm-format=s16le sample-rate=8000 \
  !  audioconvert ! audioresample ! autoaudiosink
```

Its all static though

QuantumEntangledAndy:
Sigh.. this is a horrible format to work with

QuantumEntangledAndy:
I'm thinking we should make the audio for adpcm in neolink fakesink as in no audio until we can work this out in another PR

QuantumEntangledAndy:
I think I need to find the format specification file and see why we get `Invalid data found when processing input` when doing `-acodec adpcm_ima_wav dump.adpcm` without the `-f`

QuantumEntangledAndy:
I think perhaps your `download_BC_raw.audio` has the traling 8byte pad zeros and mine has the extra sub header info of channel numbers etc but I want to find the spec to be sure

twistedddx:
download_bw_raw.audio is a Dialogic ADPCM/VOX file as far as I know.
Commonly it would have a .vox file extension but having that file extension does not improve ffmpeg's detection of the file.

twistedddx:
I think the problem is avdec_adpcm_ima_oki has no map defined for gst-libav
That unknown/unknown sink means unusable decoder.

https://github.com/GStreamer/gst-libav/blob/master/ext/libav/gstavcodecmap.c

QuantumEntangledAndy:
I was wondering why it was unknown/unknown in the [docs](https://gstreamer.freedesktop.org/documentation/libav/avdec_adpcm_ima_oki.html?gi-language=c) but why incude it in the plugin set then... sigh...

QuantumEntangledAndy:
Heres some [specifications](https://www.cs.columbia.edu/~hgs/audio/dvi/IMA_ADPCM.pdf) but its quite an old version hopefully can find a match

QuantumEntangledAndy:
Ah this is very handy [site](https://wiki.multimedia.cx/index.php/Category:ADPCM_Audio_Codecs) it has format for quite a few codecs

QuantumEntangledAndy:
FYI: The 0x0000010 byte in download_BC_raw.audio corresponds to the 0x0009ff0 in dump.adpcm

QuantumEntangledAndy:
So the first few bytes of current and index are supposed to be there I think from [this](https://stackoverflow.com/questions/55442623/adpcm-decoding-in-c)

twistedddx:
hxd
Yea same data except you have the extra 8 bytes at the start of each frame.

I do not overrun by 8 bytes, which is actually unexpected but yay for my old code getting that right.
We stop at the same point per frame.

QuantumEntangledAndy:
It seems they are wrapping the adpcm in a wav where Each first 4 bytes of block are ADPCM state

QuantumEntangledAndy:
Yes we stop at the same point because all the audio data seems to aways be a multiple of 8 bytes so no padding is needed

twistedddx:
I'm not so interested in the current/step_index bytes as the audio is usable without the decoder ever seeing it.
I think the Initial Conditions sections of this pdf may be saying the decoder recovers from inaccurate init variables after 48 samples anyway.
https://multimedia.cx/mirror/dialogic-adpcm.pdf

QuantumEntangledAndy:
Yes I'd expect some good error recovery method in a differential method otherwise errors would propogate forever. 

I'm not too interested in the index either. What I am interested in is coaxing it into a format gst knows. And it has wavparse I was hoping it would parse with the right format and we could use it.

twistedddx:
I take back my claim maybe the decoder recovers after 48 samples after re-reading that dialogic-adpcm.pdf .
It doesn't sound like that at all.
Maybe my Rick Astley sample files could be sounding better than my ffmpeg one liner is currently doing. But really as the samples show the audio is usable.

QuantumEntangledAndy:
But we can't use ffmpeg directly in gst without a lot of work. We need to coax it into gstreamers format. There is probably a way to do it in gst but I've yet to find it.

QuantumEntangledAndy:
If you have gstreamer installed then you probably have gst-launch installed too.

If you can work out a launch that plays in there then we can use it. 

```
gst-launch-1.0 filesrc location=dump.adpcm  ! rawaudioparse format=pcm pcm-format=s16le sample-rate=8000 \
  !  audioconvert ! audioresample ! autoaudiosink
```

The last bit `!  audioconvert ! audioresample ! autoaudiosink` is just to play it so we just need to tweak `rawaudioparse format=pcm pcm-format=s16le sample-rate=8000` to something that parses ADPCM.

twistedddx:
Yea I played with gst-launch-1.0 all afternoon and came up with nothing.
The rawaudioparse bit might be fine as the adpcm decoder might be able to just get the s16 and break it up as needed, although the sample count at that conversion will be wrong.
I think the next step is to give avdec_adpcm_ima_oki a mapping for audio/x-raw.

But last time I tried to compile ffmpeg under Windows I didn't get far after a day of trying.

QuantumEntangledAndy:
@twistedddx Here's something useful, put this byte sequence

```
52494646 A4910000 57415645 666D7420 10000000 10000100 401F0000 A00F0000 0A000400 64617461 50910000
```
into the beginning of your `download_BC_raw.audio` and it will play in ffmpeg without any extra options

```
ffplay download_BC_raw.audio
```

QuantumEntangledAndy:
I have disabled adpcm output for now. When this is merged I will create a tracking  issue for it.

There are two ways to get this done:

- Upstream adding support for ima_oki format into gst
- Using libav to convert it before sending it to the appsrc

twistedddx:
The wav header is nice to point players to the right codec.
VLC still can't work it out.
MPC-HC plays it fine with its directshow LAV filters that are ffmpeg based.

There is appears to be 2 format/compression codes that work.
0x10 OKI ADPCM
0x17 Dialogic OKI ADPCM

I tested a bunch of other ADPCM format codes and none gave anything useful. Eg 0x02, 0x11, 0x16, 0x36, 0x40, 0x45, 0x64 are all bad.
This makes me more confident 01wb is actual OKI ADPCM data.
I believe the 03wb is going to be IMA ADPCM, but I don't own any cameras that would generate that.

QuantumEntangledAndy:
Yes this looks like we aren't going to be able to support adpcm just yet, aac is fine.

p.s. I also got YAMAHA 0x20 to work

QuantumEntangledAndy:
But yes oki is quite a distinctive format, not many things will support it I think, the data is 4bit encoded (half a byte per sample!) with a smaller headspace than the standard adpcm

QuantumEntangledAndy:
Also adpcm_ima_ws (westwood format works too)

QuantumEntangledAndy:
Also the `avdec_adpcm_ima_*`  fail even with valid inputs which I made with ffmpeg. Only the `adpcmdec` element works correctly and that only supports microsoft and dvi format.

To get adpcm we will need to decode it to pcm using another module, we could use libsox (adding a dependency) or try binding to ffmpeg (complicated but possible) or write it ourselves

QuantumEntangledAndy:
Either way I think it should be left to another day and we just move in the aac and the scaffolding to get the adpcm to work later

QuantumEntangledAndy:
Possible could use the ffmpeg rust binding [here](https://lib.rs/crates/ffmpeg)

twistedddx:
sox knows what download_BC_raw.audio is simply by renaming to .vox
sox in_raw_adpcm.vox output_wav_pcm.wav

I always turn to sox whenever I need a-law or u-law wav files for PABX's.

Dialogic VOX isn't even that exotic, it is in all kinds of old voicemail systems.
But poor support for telecommunications media codecs(eg g711) in PC media tools is nothing new.
Sox is the only app that seems to consider them a high priority and handles them as nicely as others treat mp3 and aac.

QuantumEntangledAndy:
Yes there's an unsafe rust binding to libsox and it may be possible to use that. But I am not sure if we want to add it as a dependency, sox hasn't been maintained from what I can see for quite awhile.

QuantumEntangledAndy:
I actually started writing up something in libsox but I won't have time to finish it until maybe next week. It mostly just uses open_mem_read followed by setting up the conversions and then open_mem_write. It's just setting up the mem streams and conversions is complicated there are a ton of settings and structures to init.

QuantumEntangledAndy:
@twistedddx If you have time maybe you could try my  QuantumEntangledAndy/neolink  adpcm branch, it requires libsox to be installed.

twistedddx:
Wow that was a quick week!
I had been compiling neolink under Windows and foolishly went down the path of getting libsox to compile with MSVC 2019.
But I just realized I am probably better off using Windows Subsystem for Linux and compile neolink and libsox things from there.
I will swap to WSL tomorrow night and see if I can get things to compile.

QuantumEntangledAndy:
Yeah I have trouble stopping when I get an idea. If your on windows I recommend chocolaty to install all dependencies. I am going to try to get it to link statically too so the end user dosent need to have it.

QuantumEntangledAndy:
@twistedddx Could you maybe try the windows installer here: https://sourceforge.net/projects/sox/files/sox/14.4.2/sox-14.4.2-win32.exe/download I am hoping it sets up the dll in the right place

QuantumEntangledAndy:
So I have a static version for mac os which is nice and I think I can do linux too although because of apt/yum etc theres no need to. The issue I have is windows. The chocolatly package is exe only not the lib, the precompiled version on libsox page has the lib and all dependencies so we could package it up with the neolink and distribute it together

QuantumEntangledAndy:
@twistedddx  please check out my latest no dependency version

So.. I was having lots of dependency issues just to get one algorithm in rust.... So I wrote it myself... tada rust adpcm oki -> pcm parser

Again its a force push so please delete my branch and repull it

```bash
git checkout master
git -D adpcm
git checkout -t origin adpcm
```

twistedddx:
[dumppcm.zip](https://github.com/thirtythreeforty/neolink/files/5094950/dumppcm.zip)

Wow you are awesome! Is there some link I can click to give you money towards some drinks of your choice? 

Something isn't right though.
Attached is the output from 
Some(StreamFormat::ADPCM) => "! queue ! rawaudioparse format=0 num-channels=1 sample-rate=8000 ! filesink location=dump.pcm",

While there does appear to be real order mixed in looking at the raw data in hex there is a uniformity to the data that is clearly wrong.

Playback test via:
ffplay -f s16le -ar 8000 -ac 1 -i dump.pcm

Just for completeness the same type of data comes out from:
Some(StreamFormat::ADPCM) => "! queue ! filesink location=dump.pcm",

QuantumEntangledAndy:
I see it works for the download_BC_raw.audio (I have a little test app that just passes the input). I think I need to maintain the state between packets (adpcm depends upon previous state)

twistedddx:
Is that the state that is saved in the first 8 bytes?

QuantumEntangledAndy:
Hmm yes that is the state.... Try my latest commit and if it works I will use those bytes rather than saving them

twistedddx:
The dump.pcm works flawlessly now but there is still audio break ups in vlc/mpc for alaw and also aac when I use voaacenc ! rtpmp4apay name=pay1

QuantumEntangledAndy:
Awesome, this next commit uses the state in the stream. This may be a little better quality since we know the initial conditions now even if the camera sends the audio not from the start of the stream. Could you test it for me :)

Also I don't have the `voaacenc` in my gstreamer maybe it requires a depenedcy I don't have.

QuantumEntangledAndy:
If it works fine in the dump I think the issue is the interleaving with the video.

We apply timestamps to the video and audio separately based in the clock. This means when the VLC gets the data it thinks the two are not in sink and you get break ups. We have to fix #63 I think too

QuantumEntangledAndy:
If when testing you get an assertion error about ADPCM magic please let me know. It means ADPCM forms its own sub sub stream.... I hope not

twistedddx:
No errors.
VLC run time for the stream jumps around like mad. Back and fourth in time by 1-5 seconds.

MPC plays the stream with no audio issues.

QuantumEntangledAndy:
I see probs the time stamping. I will make the ADPCM code pretty (it's horrible at the moment and not worth a PR) and then get the timestamps sorted then ping you for a retest if that's ok you.

QuantumEntangledAndy:
Could you also try VLC using TCP connection (it defaults to UDP)

twistedddx:
It is falling to TCP. I haven't seen a UDP stream come from NVR yet.

VLC plays good, MPC plays audio good but still frame video and claiming framerate is 90000fps:
Some(StreamFormat::ADPCM) => "! queue ! rawaudioparse format=0 num-channels=1 sample-rate=8000 ! audioconvert ! rtpL16pay name=pay1",

MPC plays good, VLC plays video good but audio is choppy:
Some(StreamFormat::ADPCM) => "! queue ! rawaudioparse format=0 num-channels=1 sample-rate=8000 ! audioconvert ! rtpL16pay name=pay1",

twistedddx:
Just got this after a while of playing:
thread '<unnamed>' panicked at 'index out of bounds: the len is 49 but the index is 51', src\bc_protocol\adpcm.rs:45:27

QuantumEntangledAndy:
Oh there should be a check on the I index that stops that.... Ah but of course the index checks don't apply to when we initiise the variable from the saved stream header. Does your tests suggest we payload it as raw?

twistedddx:
rtppcmapay vs rtpL16pay
I don't know what is going to have better compatibility in media tools.
In my testing of 2 very powerful common players both seem to fail at one of them. But perhaps that would go away with better timestamping.

A goal should be to have the audio in a common format considering the fight you have just gone through to get it out of the OKI ADPCM format.

alaw has poor support in many basic players and tools.
alaw is another round of lossy compression which is a negative.

My feeling is the L16 likely will be easier to deal, less encoding although there is the endianness swap from S16LE to S16BE and lossless.
But I am open to anyone disputing anything I have said there.

QuantumEntangledAndy:
Since alaw is lossy perhaps we should avoid it. I'm in favour of aac if we can get a consistent get plugin on all platforms just because of the ubiquitousness of the format. 

Also please do try TCP VLC. The connection from NVR -> Neolink is always TCP as you say but Neolink -> VLC can be UDP or TCP depending on what VLC asks for.

QuantumEntangledAndy:
Let's get timestamps in then work out the best encoding.

twistedddx:
aac is lossy and costly on CPU.

QuantumEntangledAndy:
I'm ok with PCM too. The big endianness is just because of network order I think and it should get decoded into little endianness at the other end. I really don't mind the format since it's not me that's using it, what ever gets it working well for you is good

twistedddx:
Hah I looked at Wireshark and saw the TCP and forgot there is another stream happening on my local machines loopback.

VLC behaves exactly the same streaming via TCP or UDP.
I forced VLC to use TCP under the RTSP demuxer settings and proved via Wireshark VLC was using TCP to Neolink on my loopback adapter.

twistedddx:
Would it be worthwhile for the adpcm to be written out as s16be and therefore gstreamer would not need the audioconvert to do the endianness swap?

QuantumEntangledAndy:
I could do it like that but it would make it more difficult to proccess if we later needed to debug the raw dump. It is also a micro optimisation and likely not worth it.

twistedddx:
OK leave as is.
Also fyi you mentioned the other day the s16be might arrive at the player as le.
Vlc at least reports it is playing s16be content.

QuantumEntangledAndy:
@twistedddx I am ready for a retest from my timestamps branch please

```bash
git checkout -t origin timestamps
```

QuantumEntangledAndy:
If it still dosen't work I shall apply [this](https://gstreamer.freedesktop.org/documentation/audiorate/index.html?gi-language=c#audiorate) which should correct the audio data so that it has the same size as indicated by the timestamps.

FYI all time stamps are stored on the video frames and we assume that the audio and video streams perfectly match up time wise. 

If this is not the case the [audiorate](https://gstreamer.freedesktop.org/documentation/audiorate/index.html?gi-language=c#audiorate) plugin will apply stretching/dropping of audio such that they are the same.

QuantumEntangledAndy:
I've gone and added the audiorate plugin, it seems to be designed for this sort of issue where we don't have time stamps on both streams.

twistedddx:
I compiled the timestamp branch but things don't work for adpcm or the aac cameras.
Both give the same result of a frame or 2 of video and audio and then it is still video frame and no audio.

VLC claims it is still demuxer data in its statistics.

QuantumEntangledAndy:
Hmm perhaps it is in microseconds rather than nano I will check when I get home, doing it over a ssh tunnel is just difficult when your trying to time things.

QuantumEntangledAndy:
I've pushed a commit for micro seconds you can test it now if you want or just wait for me to do it when home.

twistedddx:
Here is what I find nano time looks like in a old test file:

https://pastebin.com/mNtJvmT0

My output of nano time is based on this C#:
uint iNanoSeconds = BitConverter.ToUInt32(pBuffer, 16);
LogMessage("Nano time:", iNanoSeconds.ToString());

QuantumEntangledAndy:
From your data the third digit


19*3*212000

Remains constant for about 30 frames.

- If it was ns then:

30 frames would be represent 0.001 s

- If it was $\mu$s then:

30 frames would represent 1 second

- If it was ms then:

30 frames would represent 1000s


So probably micro seconds. 

However I don't think this is the end of the story. I suspect this number represents the presentation time stamp. To convert from this timestamp to s we are supposed to divide by the timebase. There should be somewhere (maybe in the info header) where the timebase is stored.

twistedddx:
time_t is in the 32 byte I frame header

twistedddx:
There is often data in the "unknown" areas of I frames in particular also.

Eg
Sample 1 file:

30 30 64 63 48 32 36 34 9A C9 07 00 08 00 00 00
C0 3D 35 A4 23 00 00 00 44 9F 6D 5D 29 00 C3 00

twistedddx:
Oh Info header has what I called "camera type" also. All I know about that field is that it is different for different model cameras.

QuantumEntangledAndy:
My Info Header

- Camera set to 15 fps
```
31, 30, 30, 32, 20, 00, 00, 00, 80, 03, 00, 00, 00, 02, 00, 00,
01, 0f, 78, 08, 15, 07, 30, 0f, 78, 08, 15, 07, 30, 0f, 00, 00
```

- Camera set to 8 fps
```
31, 30, 30, 32, 20, 00, 00, 00, 80, 03, 00, 00, 00, 02, 00, 00
01, 08, 78, 08, 15, 07, 30, 22, 78, 08, 15, 07, 30, 22, 00, 00
```

- FPS is here:
```
XX, XX, XX, XX, XX, XX, XX, XX, XX, XX, XX, XX, XX, XX, XX, XX,
XX, FPS, XX, XX, XX, XX, XX, XX, XX, XX, XX, XX, XX, XX, XX, XX,
```

QuantumEntangledAndy:
Max bitrate

- Camera set to 512 max bps
```
31, 30, 30, 32, 20, 00, 00, 00, 80, 03, 00, 00, 00, 02, 00, 00,
01, 0f, 78, 08, 15, 07, 30, 0f, 78, 08, 15, 07, 30, 0f, 00, 00
```

- Camera set to 256 bps
```
31, 30, 30, 32, 20, 00, 00, 00, 80, 03, 00, 00, 00, 02, 00, 00, 
01, 0f, 78, 08, 15, 07, 36, 35, 78, 08, 15, 07, 36, 35, 00, 00
```

- Max BPS Not in header.

QuantumEntangledAndy:
So your two bytes camera type seems to be
1 byte Unknown (01)
1 bytes FPS

QuantumEntangledAndy:
@twistedddx Can you confirm this is the fps on your cameras? I could use this to work out the baserate if nothing else works

twistedddx:
Yep now that you point that out that is exactly what I see

31 30 30 31 20 00 00 00 00 0a 00 00 a0 05 00 00
00 1e 78 08 14 10 19 24 78 08 14 10 1e 01 00 00 

camera was set for 30fps.

QuantumEntangledAndy:
Ok now for a bit of maths

- Get data for 
  - Frame0
  - Frame30 (30 frames after frame0)
- Do math
(Frame30.timestamp - Frame0.timestamp) * 30

On my camera this comes out at 130,000,000 (but using 15 coz im on 15fps). I am hoping they come out as a constant and we can use that. Else I will have to keep looking

QuantumEntangledAndy:
Could you also tell me what the last 4byte of your IFrame comes out as, this is the closest number I can find to 13000000 on my stream

twistedddx:
(194212000 - 193212000) * 30 = 30,000,000

twistedddx:
Last 4 bytes are 29 01 00 00 = 297

Those last 4 bytes of I frames somewhat appear to behave like each byte is a single parameter each with the 4th byte never actually used.

QuantumEntangledAndy:
Interesting but not helpful. I shall keep looking.

QuantumEntangledAndy:
We can calculate the baserate from the FPS and the time stamps between say 60 frames but I'd rather find it in the header info

twistedddx:
It is 1,000,000 units a second for me. And not far off for you.
That seems like it is already a unit of time.

QuantumEntangledAndy:
Yes if only mine were the same then I'd use that

QuantumEntangledAndy:
Mine is a rather odd 866000 units a second

twistedddx:
Have you got other 15 frame gaps to count and see if it is consistent?

QuantumEntangledAndy:
Yeah I have an excel sheet, also I was reading the wrong column in my excel sheel it is `999000`

QuantumEntangledAndy:
It is a 1000 off from your 1 million.... Perhaps we can just use that.... I will see what sort of errors I get

QuantumEntangledAndy:
If i average the base rate over about 200 frames I get  `993421.053`

twistedddx:
If it is 1,000,000 units a second will it not reset every 10000 seconds?

QuantumEntangledAndy:
No its just a scaling factor from timestamp to seconds, no resets built in

QuantumEntangledAndy:
It could be my camera is not exactly 15fps, I will change it in the setting to something else and see how it responds

QuantumEntangledAndy:
Nope even at 8fps it is  999000 units per second

twistedddx:
Can it be a 50hz(you) vs 60hz(me) error in hardware?

twistedddx:
You have that 01 in your 1001.
Maybe that is your 1000 error?

QuantumEntangledAndy:
Hmm i think it is just rounding error. 

- At 15fps
   The difference between two frames is  66000
   If the base rate is 1000000 then the fps is 1000000/66000=15 exactly
- At 12fps
   The difference between two frames is  83000
   If the base rate is 1000000 then the fps is 1000000/83000 = 12 exactly
- At 8fps
   The difference between two frames is  125000
   If the base rate is 1000000 then the fps is 1000000/125000 = 8 exactly

QuantumEntangledAndy:
The issue is that sometimes it is  124000 between frames (8fps)

Also if I set it to something odd like 10 fps it is is a mixture between steps of  125000 and 83000, which I assume is so that on average it is 10

QuantumEntangledAndy:
So I am going to go with 1,000,000 which means.... just use microseconds....

QuantumEntangledAndy:
I think my camera is using adaptive fps rather than fixed so it can vary somewhat

twistedddx:
Also note the framerate changes when it goes to night mode.
It typical halves in night mode.
Eg 25fps becomes 12fps

QuantumEntangledAndy:
Well as long as 1000000 remains fixed we don't need to worry about it and it should work regardless of the fps, I just need to figure out why its not playing with the DTS and PTS buffers set...

QuantumEntangledAndy:
Ok so I have an answer... The player thinks the frames are in the future... We send out frames with high pts which calculate as being future frames. 

If I add a base_time subtraction when the stream starts it will play but.... I need to subtract it differently for each client that connects and reset it for a new client... There must be an easier way

QuantumEntangledAndy:
Anyways won't be able to work on it till this evening now,

QuantumEntangledAndy:
@twistedddx So I had some time over lunch to have a go and it wasn't too difficult, please try the latest of the timestamp branch

twistedddx:
I had to stick these in and let pts = match (timestamp, basetime) { to fix compile errors:

            let timestamp = self.timestamp; // Copy now before we do any mutable borrows
            let basetime = self.basetime; // Copy now before we do any mutable borrows

VLC plays video good, but adpcm audio is choppy, aac is fine
MPC-HC plays the same audio as VLC but doesn't want to play video at all.

I'm wondering if the audio is the wrong length. Perhaps there is a few more bytes of non audio data in the adpcm frame or something similar?

QuantumEntangledAndy:
Oh maybe I forgot to push a commit because it was compiling for me. I shall investigate. 

We are using the audiorate element in gst so it will duplicate or drop samples so that it matches the timestamps length and we pass the same timestamps to audio and video. So probably not lost data.

QuantumEntangledAndy:
So looking around it seems that there can be issues with sync when using two rtp payloads. The recommended way around it in gst seems to be to mux it into one stream and send them together. I will investigate this option later today

QuantumEntangledAndy:
Ok so I have it working as a mux on my h264/aac. However I am not sure what to do about  the pcm. The muxing will only work with:

audio/mpeg (1,2 4):
audio/x-lpcm:
audio/x-ac3:
audio/x-dts:
audio/x-opus:

Do you have any preference on what to convert to?

QuantumEntangledAndy:
I've done aac for now let see if it works and after that we can swap it to something else if you'd prefer

twistedddx:
faac doesn't want to work.

Your line or even this gives internal stream data error.
rawaudioparse format=pcm pcm-format=s16le num-channels=1 sample-rate=8000 ! audiorate ! faac

basesrc gstbasesrc.c:3072:gst_base_src_loop:<audsrc> error: Internal data stream error.


Both of these have the same choppy audio as before.
rawaudioparse format=pcm pcm-format=s16le num-channels=1 sample-rate=8000 ! audiorate ! voaacenc
and
rawaudioparse format=pcm pcm-format=s16le num-channels=1 sample-rate=8000 ! audiorate ! opusenc

QuantumEntangledAndy:
I see I tested it with this line and faac worked for me
```
gst-launch-1.0 -vv filesrc location=testoutself.wav !  rawaudioparse format=0 num-channels=1 sample-rate=8000 ! audiorate ! audioresample ! audio/x-raw,rate=16000 ! faac bitrate=48000 ! audio/mpeg,stream-format=adts  ! aacparse ! filesink location=gst.aac
```
The only difference is a fixed bitrate...

QuantumEntangledAndy:
Could you replace the line
```
"mpegtsmux name=mux ! rtpmp2tpay name=pay0",
```
with
```
"mpegtsmux name=mux ! filesink location=dump.ts",
```
and try and play locally

twistedddx:
rtspclient rtsp-client.c:2704:handle_setup_request: client 00000299FADC6380: no control in path '/test1'

QuantumEntangledAndy:
Is `/test1` your path? Why the leading `/`?

QuantumEntangledAndy:
`no control in path '/test1'` will be difficult to debug or understand without me having live data, it may just be an encoding issue

twistedddx:
In my config:
name = "test1"

URL I open:
rtsp://127.0.01:8554/test1


BTW:
I think this(and its linked bug report) describes the problem with rawaudioparse on live streams:
http://gstreamer-devel.966125.n4.nabble.com/RTP-audio-packets-are-sent-as-a-burst-not-in-real-time-with-rawaudioparse-td4682828.html

QuantumEntangledAndy:
I see, I could stick a wave header in front and then use waveparse like they did... This is a right rats nest

twistedddx:
Media tools only work with mp3 and aac everything else is broken.

QuantumEntangledAndy:
So I now have a wav header... Not sure if I am doing it right but we shall see

QuantumEntangledAndy:
What do you mean by `MediaTools` I don't know what this is?

QuantumEntangledAndy:
Please let me know if the wave header helps

twistedddx:
It was just an observation of media applications and media devices in general.
I find there is good support for mp3 and aac(if inside mp4) and poor support for basically anything else.

Gstreamer appears to be littered with buggy code such as the libav_adpcm decoders and pcm parser.
It is impossible to support all exotic formats but there is a very limited amount of codecs that get good support.

QuantumEntangledAndy:
Oh I see what you mean. I've been leaning that the hard way what with the fun of the ADPCM parser...

QuantumEntangledAndy:
Currently I add the wav header to every packet. But depending on how that works for you I may need to add it only at the first packet.

QuantumEntangledAndy:
If it still won't work after that I have a plan to directly set the caps rather than parsing.

twistedddx:
faac is still broken but opusenc is working although there is some additional noise in the audio.
Additional noise might just be from the fast opus encode.
Maybe we can go back to not muxing?

QuantumEntangledAndy:
Sooo.. I finally got my method to decode in the same way as their....


This log shows how thiers and mine decodes each block. 
```
Theirs: -9401:81
[2020-08-24T04:17:17Z INFO  rust_wav_test::adpcm_self] Expected: -9401, Got: -9401, Factor: 1, Diff: 0
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 1:  1889:82
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 2:  9247:83
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 238:  2524:73
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 239:  653:71
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 240:  1757:69
Theirs: 1757:69
[2020-08-24T04:17:17Z INFO  rust_wav_test::adpcm_self] Expected: 1757, Got: 1757, Factor: 1, Diff: 0
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 1:  -9382:76
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 2:  -758:79
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 238:  5851:76
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 239:  -3996:74
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 240:  613:72
```

and this is mine

```
Mine: -9401:81
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 1:  1889:82
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 2:  9247:83
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 238:  2524:73
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 239:  653:71
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 240:  1757:69
Mine: 1757:69
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 1:  -9382:76
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 2:  -758:79
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 238:  5851:76
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 239:  -3996:74
[2020-08-24T04:17:17Z DEBUG rust_wav_test::adpcm_self]     During 240:  613:72
```


... But check what it sounds like. It is all static!

[using_bc_header_vars.pcm.zip](https://github.com/thirtythreeforty/neolink/files/5115754/using_bc_header_vars.pcm.zip)
[using_own_vars.pcm.zip](https://github.com/thirtythreeforty/neolink/files/5115755/using_own_vars.pcm.zip)

QuantumEntangledAndy:
I think what is happening is that the encoder uses the approximate form and therefore stamps the block header with the approximate results, but the decoder on the NVR or smartphone uses the accurate form (which sounds okish). 

I think the way forward is to use the full non-approximate form, and only insitialising the very first part of the stream, at camera connect but not at each block, with the approximate form in the header

QuantumEntangledAndy:
I was hoping that we could use the block header and not have to maintain a state variable... why bc

QuantumEntangledAndy:
Maybe this is the reason the encoder was dropped for aac?

QuantumEntangledAndy:
@twistedddx Whenever you have time please checkout the latest on the caps branch.

twistedddx:
Both of those latest zip files play nicely for me.

Thats sounds correct:
ffplay.exe -f s16le -ar 8000 -ac 1 -i using_bc_header_vars.pcm

QuantumEntangledAndy:
Ah I see I was using the -acodec pcm_u16le codex... not the signed version haha

QuantumEntangledAndy:
Ok then I can use the internal state YAY

QuantumEntangledAndy:
Latest on the caps uses the approximate method and the state variables in the DVI4 header, please check it out :)

twistedddx:
Sorry I got sidetracked trying to see what kind of difference OKI method vs DVI method does for the file.
For quiet spots the DVI stays quiet while the OKI behaves like it is massively volume boosted.
In reality quiet spots being volume boosted is not bad for CCTV.

Once the music starts I can't even pick the 2 files apart in foobar's foo_abx plugin.

DVI wav created from your using_bc_header_vars.pcm
ffmpeg.exe -f s16le -ar 8000 -ac 1 -i using_bc_header_vars.pcm dvi.wav

OKI wav created from original dump.adpcm with 8 byte headers removed.
ffmpeg.exe -f s16le -ar 8000 -ac 1 -acodec adpcm_ima_oki -i dump_clean.adpcm oki.wav
[test_abx.zip](https://github.com/thirtythreeforty/neolink/files/5116292/test_abx.zip)

I shall get to caps testing now!

QuantumEntangledAndy:
OKI has fewer quantiziation levels which means lower resolution, it cannot handle rapid changes in frequency but for things like speech which has only fairly slow changes in frequency it should be fine. In most cases I don't imagine we can distingish much between them.

QuantumEntangledAndy:
OKi volume boosting is likely because it is scaled from a 12 bit integer to a 16 bit integer as part of the decoding proccess.

twistedddx:
[dump4.zip](https://github.com/thirtythreeforty/neolink/files/5116440/dump4.zip)
 I get popping out the players which is also clear in the dump

Dump capture with:
Some(StreamFormat::ADPCM) => "! filesink location=dump.pcm",

QuantumEntangledAndy:
Yeah copy pasta bug.. Which iv'e hopefully fixed. I read the last sample from the header as u16 instead of i16.

QuantumEntangledAndy:
Please try now

twistedddx:
All works great now!

my config for testing:
bind = "0.0.0.0"
# bind_port = 8554
#Reolink RCL-410 4mp H264/AAC
[[cameras]]
name = "test0"
username = "xxx"
password = "xxx"
address = "192.168.1.2:9000"
stream = "mainStream"
channel_id = 0

#Swann HND-818 4mp H264/ADPCM
[[cameras]] 
name = "test1"
username = "xxx"
password = "xxx"
address = "192.168.1.2:9000"
stream = "mainStream"
channel_id = 1

I can connect to both cameras at the same time from either VLC or MPC-HC.

Things of note but not serious
-MPC-HC almost always reports the FPS of the video for either type of camera incorrectly but it plays ok.
Actual 30 FPS but it often claims 90090.094fps or something between 30-100fps.
-MPC-HC can sometimes connect audio only
-VLC takes several seconds for AAC audio to stabilize. At the start there is several audio drops.

Both caps and rawaudioparse methods now behave exactly the same in my tests. Both work fine.
Some(StreamFormat::ADPCM) => "caps=audio/x-raw,format=S16LE,layout=interleaved,channels=1,rate=8000,channel-mask=(bitmask)0x0  ! queue silent=true max-size-bytes=10485760  min-threshold-bytes=1024 ! audioconvert ! rtpL16pay name=pay1", 

Some(StreamFormat::ADPCM) => "! queue silent=true max-size-bytes=10485760  min-threshold-bytes=1024 ! rawaudioparse format=pcm pcm-format=s16le sample-rate=8000 num-channels=1 interleaved=true ! audioconvert ! rtpL16pay name=pay1",

QuantumEntangledAndy:
I'd like to try something else while we are here: it is on the `adpcm_gst` branch

twistedddx:
....it works....

But I think your decoder sounds better. This sounds comes out a bit clicky.
Dump from:
Some(StreamFormat::ADPCM) => "caps=audio/x-adpcm,layout=dvi,block_align=244,channels=1,rate=8000  ! queue silent=true max-size-bytes=10485760  min-threshold-bytes=1024 ! adpcmdec ! filesink location=dump.pcm",

[dump5.zip](https://github.com/thirtythreeforty/neolink/files/5116751/dump5.zip)

QuantumEntangledAndy:
I've put all of our work into this PR can you get the PR version `detect` branch and make sure everything is still working.

twistedddx:
Looking good and test good for VLC and MPC-HC for AAC and ADPCM cameras.

I did get a warning on compile:
warning: unused import: `RTSPPublishClockMode`
  --> src\gst.rs:12:33
   |
12 |     RTSPAuth, RTSPMediaFactory, RTSPPublishClockMode, RTSPServer as GstRTSPServer, RTSPToken,
   |                                 ^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: 1 warning emitted




Also is you OKI and IMA max_sample_size consistent?
It looks like IMA should be set as 32768 ?

QuantumEntangledAndy:
Yeah I have already pushed, to remove that warning. (You can pull it if you want but it was a force push so maybe best to delete and recheckout the branch).

I will add the change for the IMA to 32768 seems I missed that bit.

QuantumEntangledAndy:
Ok commit for the IMA max size is done. It is also working fine for my E1.

QuantumEntangledAndy:
@thirtythreeforty I think we finally have working audio for all camera. Please come and review when you are free :).

QuantumEntangledAndy:
Summary:

- Add a structure for GstOutputs, this:
  - Controls the gst launch string
  - Has references to both the audio and video appsrcs (we need a new appsrc for each one)
- Adds a ADPCM decoder.
  - We discovered right at the end of all of our issues that we can also decode in gst, if we strip the first 4 bytes of the block header and force a set caps as DVI4. However @twistedddx thought that our own decoder has better performance
- Strips the rx_timeout from media_packet and uses your constant defined elsewhere as RX_TIMEOUT (I missed that I could do that when I merged in your time branch)
- Rust fmted

twistedddx:
I am certain that the 7A in the magic is 122 s16's aka 244 bytes which represents the DVI4 data payload.
That is 480 samples per frame.

QCam have a  00 01 52 00 which is 82 s16's aka 164 byte DVI4 data payload.
That is 320 samples per frame.

HiSilicon claim "Generally, the frame length is a multiple of 80, namely, 80, 160, 240, 320, or 480"

QuantumEntangledAndy:
Haha I was just thinking about the beading being half the block size and came to ask you about it only it see your own post.

QuantumEntangledAndy:
I might add a commit that checks the block size is valid.

QuantumEntangledAndy:
I wonder if  00 01  is the number of channels

QuantumEntangledAndy:
So on the assumption it is block size and number of channels (it should be as thats the data required to decode DVI) I have added a commit that reflects this. Please test on your cameras.

twistedddx:
I thought that once but I am fairly confident that also comes from HiSilicons VoiceSDK.
When it creates data it tags it with 01 for Hisilicon voice frame and with the s16 audio data length.
I believe both these bits of data are coming from that HiSilicon lib.

twistedddx:
hi_voiceengine
Here is the doc that talks about the length in s16 and the 01.

QuantumEntangledAndy:
Yes I remember reading that part. Just at the time I couldn't understand the length of speech data part back then.  It only dawned on me earlier when I was considering the DVI block size that it was half the size. Why is it half I have no idea. Perhaps so that it fits in one byte even when it's > 255? That may be true if HISilicon are assuming that the largest block size is 480.

QuantumEntangledAndy:
Anyways I have changed the ADPCM decoder to try and explain this part of the code. I also assert the number of bytes is a multiple of the (data_frame_size+4) and then I chunk the bytes into chunks of length (data_frame_size+4) so that even if we get more than one frame in an audio packet it will decode correctly.

twistedddx:
The VoiceSDK deals with adpcm, g711 and pcm. Maybe the s16 size simply came from 16bit PCM side of things and is more a quirk when dealing with compressed formats.
The 01 could even be the frame count?
If it was 00 02 7A 00 would you expect a double 244 byte DVI payload back to back?

QuantumEntangledAndy:
It's all unknown until we see the behaviour live. For now I have an assertion that it's 0x0001 so at least if it is something new it will crash and hopefully people will shout about it and let us know.

thirtythreeforty:
> @thirtythreeforty I think we finally have working audio for all camera. Please come and review when you are free :).

I plan to! This is quite the thread so I may have to do it in pieces.

One thing I'd like to go ahead and bring up is your comment:

> at least if it is something new it will crash and hopefully people will shout about it and let us know

This worries me - it's a security camera setup and it needs to _not crash_ if at all possible... It would be much better to loudly log an error, let the audio stream corrupt, and keep the video threads running.

thirtythreeforty:
Also, again, huge thank you for working this out across all the bizarre audio formats. I'm looking forward to landing the support!

QuantumEntangledAndy:
> > @thirtythreeforty I think we finally have working audio for all camera. Please come and review when you are free :).
> 
> I plan to! This is quite the thread so I may have to do it in pieces.
> 
> One thing I'd like to go ahead and bring up is your comment:
> 
> > at least if it is something new it will crash and hopefully people will shout about it and let us know
> 
> This worries me - it's a security camera setup and it needs to _not crash_ if at all possible... It would be much better to loudly log an error, let the audio stream corrupt, and keep the video threads running.

Ok so I have swapped to error logging instead :)

QuantumEntangledAndy:
@twistedddx  Are you refering to my change from 

```rust
channel_id: channel_id
```

to 

```rust
channel_id
```

This change won't effect its operation, rust can implicitly assign variable to key if key and varaiable have the same name. I did this to conform with `clippy` styling rule [`redundant_field_names`](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_field_names).

Sorry for late reply completely missed this message while working on other things

QuantumEntangledAndy:
@thirtythreeforty So I think I understand why you were using `Write` as the input for the start_video (its supposed to be a generic library function), if you want I can change the currently implementation to something generic.

Sorry, just a thought while I was writing up the `start_motion` part for the mqtt code.

src/bc_protocol/adpcm.rs
```diff
@@ -0,0 +1,226 @@
...
+                // We also cast to i16 at this point ready for the conversion to u8 bytes of the output
+                let scaled_sample = (sample as isize * (i16::MAX as isize)
+                    / (context.max_sample_size - 1) as isize)
+                    as i16;
+
+                // Get the results in bytes
+                result.extend(scaled_sample.to_le_bytes().iter());
+
+                // Increment the step index
+                step_index = step_index as isize + context.changes[unibble];
```
thirtythreeforty, line 218: This slice is okay, because you've constrained the index to be 0..16 by dint of the nibble bitmask.

src/bc_protocol/media_packet.rs
```diff
@@ -129,7 +154,7 @@ impl MediaData {
             MAGIC_IFRAME => MediaDataKind::VideoDataIframe,
             MAGIC_PFRAME => MediaDataKind::VideoDataPframe,
             _ => {
-                trace!("Unknown magic kind: {:x?}", &magic);
+                //trace!("Unknown magic kind: {:x?}", &magic);
```
thirtythreeforty, line 157: Why comment this out?

src/bc_protocol/media_packet.rs
```diff
@@ -138,6 +163,39 @@ impl MediaData {
...
+
+    pub fn media_format(&self) -> Option<StreamFormat> {
+        let kind = self.kind();
+        match kind {
+            MediaDataKind::VideoDataIframe | MediaDataKind::VideoDataPframe => {
+                let stream_type = &self.data[4..8];
+                const H264_STR_UPPER: &[u8] = &[0x48, 0x32, 0x36, 0x34];
+                const H264_STR_LOWER: &[u8] = &[0x68, 0x32, 0x36, 0x34];
+                const H265_STR_UPPER: &[u8] = &[0x48, 0x32, 0x36, 0x35];
+                const H265_STR_LOWER: &[u8] = &[0x68, 0x32, 0x36, 0x35];
```
thirtythreeforty, line 175: Any reason not to spell it out as a byte literal `b"h264"`?

thirtythreeforty (changes requested):
This looks good. I have a few comments, mostly related to array slicing.

One other question, I'm starting to lose track of which cameras use which video and audio formats. If you would write down what each one uses, I can convert it to a docs/ page.

src/bc_protocol.rs
```diff
@@ -245,7 +245,7 @@ impl BcCamera {
 
     pub fn start_video(
         &self,
-        data_out: &mut dyn Write,
```
thirtythreeforty, line 248: Just caught this re: your comment. Yes, please keep as `dyn Write` (and add a second argument for the audio stream) :+1:

thirtythreeforty:
@QuantumEntangledAndy and @twistedddx I am reading backlog - lots of good discussion. I am fine with merging the Rust ADPCM decoder, however, I see you realized it can be fed to Gstreamer. I like the SLOC reduction it would bring, and I actually think Gstreamer would be more efficient (no offense to this implementation, it's just Gstreamer's wheelhouse).

Based on the comments it seems like ADPCM doesn't stream, so does Gstreamer have enough info to recover the original stream without parsing the way the Rust does it? Is there any downside to that route?

src/bc_protocol.rs
QuantumEntangledAndy: I have to also pass up that the format changes that set gst to use the right launch str. maybe via a hints callback. Would that be acceptable?

src/bc_protocol/adpcm.rs
```diff
@@ -0,0 +1,226 @@
...
+    // Unused, originally we thought BC might be using OKI but it is actually DVI4
+    fn new_oki() -> Self {
+        Self {
+            max_step_index: 48,
+            steps: vec![
+                16, 17, 19, 21, 23, 25, 28, 31, 34, 37, 41, 45, 50, 55, 60, 66, 73, 80, 88, 97,
+                107, 118, 130, 143, 157, 173, 190, 209, 230, 253, 279, 307, 337, 371, 408, 449,
+                494, 544, 598, 658, 724, 796, 876, 963, 1060, 1166, 1282, 1411, 1552,
+            ],
+            changes: vec![-1, -1, -1, -1, 2, 4, 6, 8, -1, -1, -1, -1, 2, 4, 6, 8],
```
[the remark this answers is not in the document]
QuantumEntangledAndy: Oh sure :). Rust lvl up time.

src/bc_protocol/adpcm.rs
```diff
@@ -0,0 +1,226 @@
...
+    // To solve this reolink caches the intermediate variables (step_index and last_output)
+    // Into the block header in the stream. When ever an adpcm packet arrives it starts with
+    // 0x0001 which is the frame time from HISilicon documentation (I think) following by
+    // 0xWW which is half the block size
+    // 0xYY which is the last output
+    // 0xZZ which is the step index of the last output
+    // We must initialise our decoder with this data
+
+    // Check for valid number of frame type
+    let frame_type_bytes = &bytes[0..2];
```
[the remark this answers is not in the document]
QuantumEntangledAndy: I will add some bound checks

src/bc_protocol/adpcm.rs
```diff
@@ -0,0 +1,226 @@
...
+                .expect("slice with incorrect length"),
+        ) as isize;
+        let step_index_bytes = &bytes[6..8];
+        let mut step_index = u16::from_le_bytes(
+            step_index_bytes
+                .try_into()
+                .expect("slice with incorrect length"),
+        ) as isize;
+
+        // To avoid casting to u8 <-> u16 <-> u32 and back all the time I just do all maths in u/isize
```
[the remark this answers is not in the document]
QuantumEntangledAndy: I/U32 would be sufficient. The data itself is u16. But we apply a delta to that data which can subtract and then go below zero. Therefore i need to use an I type for it during calculation. I32 is the best fit as it allow us to use the whole U16 range and go negative.

QuantumEntangledAndy:
There are three pieces of information in the ADPCM data

- 1: BC data in its header
- 2: HISilicon header which include **block size**
- 3: The adpcm state and data

Gstreamer can decode 3 if told the blocksize in 2. I would need to feed this data in which I would do in the enum.


```rust
AAC,
ADPCM(block_size)
```

QuantumEntangledAndy:
Also I do prefer gstreamer too. Not that it wasn't fun to write the decoder, but rather would rather not have to maintain it. The best advantage I think it has though it that it knows the HISilicon format and CA decode that part too.

twistedddx:
Rust adpcm also audibly sounds better than the gstreamer adpcm decoder. 
This may be accurate vs approximate decoding issue.
Gstreamer may be using the accurate decode with multiply and divide while the BC format uses bitshifts. I read the official references for the dvi4 standard are as bitshifts so the approximate calculations are in fact the official standard.

twistedddx:
For camera types:

Modern protocol with XML encryption. H265/H265 video, aac audio:
Reolink B800 4k (as seen by thirtythreeforty) 

Modern protocol with XML encryption. H264 video, aac audio:
Reolink RLC-410 4mp v1
Reolink RLC-410 4mp v2
Reolink RLC-410 5mp
Reolink RLC-520 5mp
Reolink E1 (as seen by Quantumentangledandy) 

Modern protocol without xml encryption. H264 video, dvi4(244bytes) adpcm audio:
Swann NHD-818 3mp

Legacy protocol. H264 video, no audio:
Swann NHD-805 1mp
Swann NHD-806 1mp

QuantumEntangledAndy:
> Gstreamer may be using the accurate decode with multiply and divide while the BC format uses bitshifts. I read the official references for the dvi4 standard are as bitshifts so the approximate calculations are in fact the official standard. 

I checked the source code when I was trying to see if we can use the gstreamer decoders and I can say they are also using the approximate method in a very similar fashion to our own. 

The official gstreamer one can also handle 2 channel audio. Not that we are using it. Whereas our own one cannot.

QuantumEntangledAndy:
I have a deadline at work coming up so won't be able to work on these changes for a few days. Will get back to it when I can

thirtythreeforty:
> Rust adpcm also audibly sounds better than the gstreamer adpcm decoder.
> This may be accurate vs approximate decoding issue.
> Gstreamer may be using the accurate decode with multiply and divide while the BC format uses bitshifts. I read the official references for the dvi4 standard are as bitshifts so the approximate calculations are in fact the official standard.

This is reason enough to keep it in. "More correct" will always trump "more convenient" in my mind. Again, really nice work chasing all these formats down.

thirtythreeforty:
> I have a deadline at work coming up so won't be able to work on these changes for a few days. Will get back to it when I can

No worries. Heaven knows I've been out for a while :)

thirtythreeforty:
Also @twistedddx - can I add your Swann cameras to the list of supported devices in the readme?

src/bc_protocol.rs
thirtythreeforty: Oh, good point. In that case I think the callback would probably be uglier; you can leave this. In general I would love for this to be a generic library on crates.io but that refactor is _definitely_ down the road.

twistedddx:
> Also @twistedddx - can I add your Swann cameras to the list of supported devices in the readme?

Yes.
My post listed the cameras abilities but it did lack what that means to Neolink.

Cameras using the legacy protocol such as nhd-805/nhd-806 are not supported by neolink.

Cameras using modern protocol without xml encryption such as nhd-818 depend on #79. A NVR with current firmware acting as a proxy for the camera does encrypt the XML between Neolink and NVR allowing Neolink to connect to the nhd-818 streams.

src/bc_protocol/adpcm.rs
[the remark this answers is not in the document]
QuantumEntangledAndy: Done

src/bc_protocol/adpcm.rs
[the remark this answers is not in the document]
QuantumEntangledAndy: Done

src/bc_protocol/adpcm.rs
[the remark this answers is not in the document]
QuantumEntangledAndy: This is done although it does add some casts to usize during array indexing. I was under the impression that u/i size would match the machines architecture so I would expect them to be at least i/u32 as I doubt there are any many target archs with less than 32 bits

src/bc_protocol/adpcm.rs
QuantumEntangledAndy: Yes this part here should all be in range be nature of the constraints

QuantumEntangledAndy:
I have made the request changes..... Anxiously awaiting review now....

ZeldoKavira:
I've been running this branch for about 24 hours now with 1 810A (4k@25fps) and have not had a single issue, audio works great. Thank you!

QuantumEntangledAndy:
Glad it's working for you. We should be about ready to merge to master too :)

ZeldoKavira:
@thirtythreeforty Hey George, any way you can review this PR? would love for it to hit the main branch.

thirtythreeforty (approved):
This looks great! Thank you very much for the hard work and reverse engineering.

thirtythreeforty:
@QuantumEntangledAndy I am going to squash-n-merge this if it's okay with you. If you think there's value in keeping the history, please do say so. Otherwise I'll be merging Soon(tm).

QuantumEntangledAndy:
Squashing should be ok. There's a lot of trial and error in the commits that probably don't need to be there

thirtythreeforty:
Now - time for me to implement a Windows service and I'd say this is ready for a release.

PR #69, opened 2020-08-14
Use an Arc Mutex to set the clock only once

### LINKED ISSUE #57: Check/set time only once per camera, not per stream
This sets the time twice if you have `mainStream` and `subStream` enabled. Not really the end of the world but also not tidy. One of the streams needs to be instructed to be the "management" stream that does all this housekeeping.

_Originally posted by @thirtythreeforty in https://github.com/thirtythreeforty/neolink/pull/54#issuecomment-667162987_

QuantumEntangledAndy:
Here is an attempt at setting the clock only once per camera.

It is my first foray into cross thread locking in rust so I hope I got it right. I mostly just followed the [docs](https://doc.rust-lang.org/std/sync/struct.Mutex.html) on `Arc<Mutex>`.

It works by creating a `Arc<Mutex<bool>>` in the camera config (with serde=skip so that it cannot be set in the config file). I use try_lock() to attempt to lock it and if successful it goes on to set the time, if unable to lock it just skips the time setting code entirely.

closes #57

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -6,6 +6,7 @@ use std::net::SocketAddr;
 use std::time::Duration;
 use validator::{Validate, ValidationError};
 use validator_derive::Validate;
+use std::sync::{Arc, Mutex};
 
 lazy_static! {
     static ref RE_STREAM_FORM: Regex = Regex::new(r"^([hH]26[45]|[ \t]*[!].*)$").unwrap();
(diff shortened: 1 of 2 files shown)


thirtythreeforty:
So the thing I don't like about this approach is it doesn't handle the following sequence:

- Camera boots, needs time
- Neolink connects
- Neolink sets time, once
- Camera loses power
- Neolink reconnects
- Camera does not get time set again

Instead, what if we designate one of the stream threads as the "management" thread, and it will always try to check to see if the time is set when it connects. That way, there's no race between threads, and you don't have to modify the camera config struct.

thirtythreeforty (changes requested):
Try passing a new `manage: bool` argument to `camera_main()`. You can calculate it when you're setting up the camera threads based on if you're streaming HD/SD or just SD (just pick the first you come across or whatever's easiest). That way you don't need locks at all!

QuantumEntangledAndy:
Ok, less fun without the locks :p, but here what about this one :)

thirtythreeforty (approved):
Looks good! Yeah without locks it's less fun :)

PR #70, opened 2020-08-14
Allow setting channel_id

twistedddx:
This allows setting the channel_id in the config file.

Eg:
[[cameras]]
name = "test0"
address = "192.168.1.2:9000"
channel_id = 0

[[cameras]]
name = "test2"
username = "admin"
address = "192.168.1.2:9000"
channel_id = 2

### DIFF
--- a/src/bc_protocol.rs
+++ b/src/bc_protocol.rs
@@ -244,7 +244,7 @@ impl BcCamera {
         Ok(())
     }
 
-    pub fn start_video(&self, data_out: &mut dyn Write, stream_name: &str) -> Result<Never> {
+    pub fn start_video(&self, data_out: &mut dyn Write, stream_name: &str, channel_id: u32) -> Result<Never> {
         let connection = self
             .connection
             .as_ref()
@@ -261,7 +261,7 @@ impl BcCamera {
             BcXml {
                 preview: Some(Preview {
                     version: xml_ver(),
-                    channel_id: 0,
+                    channel_id: channel_id,
                     handle: 0,
                     stream_type: stream_name.to_string(),
                 }),
--- a/src/config.rs
+++ b/src/config.rs
@@ -71,6 +71,10 @@ pub struct CameraConfig {
     pub stream: String,
 
     pub permitted_users: Option<Vec<String>>,
+
+    #[validate(range(min = 0, max = 31, message = "Invalid channel", code = "channel_id"))]
+    #[serde(default = "default_channel_id")]
+    pub channel_id: u32,
 }
 
 #[derive(Debug, Deserialize, Validate, Clone)]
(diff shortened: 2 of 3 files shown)


QuantumEntangledAndy:
Do you know what the channel id is used for?  We have an issue somewhere that some camera can't stream main and substream together and I wonder if this will help or not?

twistedddx:
For NVR's the channel_id is which camera.
8 channel NVR has channel_id 0 through 7.
Each channel has its own main and sub stream.

With this I can now stream all 8 cameras simultaneously with Neolink from a single NVR by specifying which channel_id I want per neolink camera entry.

For me this helps me play with neolink while leaving my cameras connected via the NVR.

QuantumEntangledAndy:
I see. Well your code looks nice and clean to me just have to wait for thirtythreeforty to comment.

thirtythreeforty:
Hey sorry for delay. Thanks for the patch, I've had several people asking whether they could use their existing NVR with Neolink.

Do you know if the NVR provides any named indication of channels? It's not the end of the world if you have to specify channel number. However if users could give the name of the camera then this would prevent issues with the order the cameras were registered with the NVR (which presumably is how the channel numbers are picked?).

twistedddx:
I dont know if the Nvr itself keeps track of camera names, I don't believe it does. I assume the channel_id came from analog camera dvr where it was a physical port. 

I think the nvr just links an ip or mac to a channel.

Judging by the official client and local Nvr behaviour:
The camera name is part of a cameras config.
Camera config can be downloaded per channel. 
Eg to find a camera by name you would open a channel, get the config, check the name in the config and if it isn't what you wanted you would move to the next channel.

brianhanifin:
I need this functionality. I also would like to continue using my Reolink NVR to record constantly (instead of putting that entire load on my server). While doing other stuff with the high quality feed on my server.

QuantumEntangledAndy:
I believe thirtythreeforty is still unavaible at the moment but I imagine this will get merged as soon as he's back. You can pull a copy of [twistedddx](https://github.com/twistedddx/neolink) version which has this functionality in the mean time.

brianhanifin:
> I believe thirtythreeforty is still unavaible at the moment but I imagine this will get merged as soon as he's back. You can pull a copy of [twistedddx](https://github.com/twistedddx/neolink) version which has this functionality in the mean time.

Thanks, I would love to. But I'm running it on Docker inside Unraid. If I could access the files I could make the 5 edits by hand. I'll just have to wait patiently... unless I come up with a work-around.

QuantumEntangledAndy:
I see well, I can build you the docker and publish give me about half an hour

QuantumEntangledAndy:
Ok you can try this one:

```bash
docker pull quantumentangledandy/neolink:channel_id
```

thirtythreeforty:
Hi! I'm back again :) perhaps reduced capacity for a while, but anyway things are settling down.

This looks good to me. I would like to have a better way of forwarding channels, but in the meantime this works fine. Thanks for the patch!

thirtythreeforty (approved):

twistedddx:
Worth noting that my HI3535 based device can only handle Neolink connecting to 2 concurrent mainStreams (2x 5mp tested) streams at the same time.
It can handle connecting all 8 subStreams at the same time.

PR #71, opened 2020-08-15
add tzdata to Dockerfile

m1k1o:
Timezone issues with my camera. It's always 2 hours in the past of my actual Timezone. Not sure if this mitigates this problem, but certainly will be only helpful.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,7 @@ RUN cargo build --release
 FROM docker.io/alpine:latest
 
 RUN apk add --no-cache -X http://dl-cdn.alpinelinux.org/alpine/edge/testing libgcc \
+  tzdata \
   gstreamer \
   gst-plugins-base \
   gst-plugins-good \


thirtythreeforty:
Can you post your logs surrounding when Neolink sets the camera's time? I'm curious to see what it thinks the time is, vs the system time, vs the time it sets on the camera.

m1k1o:
It seems to be settingt correct time from container. But container has wrong time, host OS has `04:00:29`.
image

QuantumEntangledAndy:
Do you set the TZ enviromental variable when starting docker? I don't think it has a means to autodetect the TZ even with the package installed.

We might also want to setup the right symlink for the TZ (not all apps read the enviromental variable) using something like this in the entry point.

```bash
if [ -n ${TZ} ] && [ -f /usr/share/zoneinfo/${TZ} ]; then
  ln -sf /usr/share/zoneinfo/${TZ}  /etc/localtime
fi
```
[source](https://github.com/sameersbn/docker-gitlab/issues/77#issuecomment-159774461)

thirtythreeforty:
@m1k1o does adding the tzdata package actually resolve the problem? I'm not sure we actually do any timezone calculations, although I admit I don't know the intricacies of whatever the standard library does.

m1k1o:
Yes, indeed, see discussion [here](https://github.com/gliderlabs/docker-alpine/issues/136). In container is now correct timezone:

```
$ docker exec neolink date; date
Sun Sep 13 11:14:53 UTC 2020
Sun Sep 13 11:14:53 CEST 2020
```

Without `tzdata`:

```
$ docker exec neolink date; date
Sun Sep 13 09:15:21 UTC 2020
Sun Sep 13 11:15:21 CEST 2020
```

Although neolink does not inherit time zone from its container:

```
$ docker logs neolink
[2020-09-13T09:14:01Z INFO  neolink] Neolink 0.3.0 (unknown commit) release
[2020-09-13T09:14:01Z INFO  neolink] cam: Connecting to camera at 192.168.3.2:9000
[2020-09-13T09:14:02Z INFO  neolink] cam: Connected and logged in
[2020-09-13T09:14:02Z INFO  neolink] cam: Camera time is already set: 2020-09-13 9:14:02 +0
[2020-09-13T09:14:02Z INFO  neolink] cam: Starting video stream mainStream
```

thirtythreeforty (approved):
Cool, thanks. Just wanted to make sure it made the difference. There are _so many_ gotchas when building Docker images.

PR #78, opened 2020-09-25
Motion Detection and MQTT

QuantumEntangledAndy:
This is more of a proposal for comments than something I want to merge just yet, but we can do motion detection event publishing using MQTT.

Currently I have:
- Gotten the MQTT server up and running
- Added the BC message ID num 33
- Tried to subscribe to both message 3 and 33

The subscription to both 3 and 33 is not quite working as expected but I am working on it 

Works towards #58

---

**Update:** This started off as just a simple motion detection and MQTT however in the process it was found that our undestanding of the protocol needed updating and that the bc_protocol and the serde needed updating. Sorry for that...

PR #80, opened 2020-10-01
Dissector update and some notes about BC message

QuantumEntangledAndy:
While working on motion detection #78 we discovered that the wireshark dissector was not able to dissect packets when there were multiple BC messages inside.

This PR reworks the dissector to support multiple BC messages per packet

I have also added a markdown file that documents the BC messages I have observed and twistedddx has added his own comments and insights into the differences of the older cameras


QuantumEntangledAndy:
I have added some documentation about the BC protocol and the media packets

twistedddx:
Nice.
You have a typo of "wither" in mediapacket.md
You refer to nonce as noonce a couple times. I have heard people describe nonce as meaning "number once" but that is not correct etymology of nonce.

QuantumEntangledAndy:
Ah yes my typos are unfortunatly already in this repo :(....  Thanks for the proof read :) I shall edit

twistedddx:
Messages that are not already in your doc file or are substantially different for me to what you have seen.
https://www.wasteofcash.com/BCConvert/messages.txt

Note I have seen the unknown byte used twice now in a response from a camera which does not have XML.
190 PTZ and 208 LedState.
In both cases the feature was unsupported by the camera.

I did not include much about the write commands. It appears you just send the XML in the same format that you receive in the paired read command.

Messages related to replay function:
https://www.wasteofcash.com/BCConvert/replay_messages.txt
The replay area looks ugly. Message  5 starts a stream of pure raw BC media with no BC protocol headers.
Some of the request packets have XML in the payload and then a slightly different version of the same data in the binary payload also.
A few of the commands somewhat look like complete duplicates of the same thing. 


Also an overview of the login sequence I see on NVR:
https://www.wasteofcash.com/BCConvert/sequence.txt

QuantumEntangledAndy:
This is quite handy reference. When I get the time I plan to make the dissector dump the data to file for me in the right format for my markdown table then just need to click all the buttons to get a complete set for my E1.

Maybe I will learn what the unknown byte with the PTZ does since my camera should support it. Unless the unknown byte is the byte for invalid or unknown header. Hmmm perhaps should send some faulty data with neolink and see how the response differs.

twistedddx:
From my testing the unknown byte may be error. 
The camera sends an ack response for the message but flags error to the client. 

I would suggest to be careful deliberately sending incorrect data to the devices. I do not believe the devices do much checking of data.

I bricked a Swann NHD-818 (IPC_388M) 2 years ago trying to flash firmware intended for Reolink RLC-410 (IPC_3816M).
The firmware update process appears to only check the file name matches for that model.  Maybe the 8M vs 16M is storage size, that would match their released firmware sizes also.
I never checked if I could recover the IPC_388M via UART.

QuantumEntangledAndy:
Good point, I have already sent an incorrect packet where I had the xml message set into the binary block.
In this case it sent back the header but without the requested data. I did not save the packet so can't check that byte.

As long as I stick to roughly the right messages I should be ok, just should avoid things like incorrect message lengths I don't know what sort of protections it has against buffer overflow etc.

QuantumEntangledAndy:
So the messages md had an overhaul today. I made a script to parse all messages from the camera directly into the md format and then just clicked away at everything the client can do (except things invloving SDD card coz I don't have one :(...).

In short this should be every message type at least basically known.

Messages id 192 is the only one I still don't know what it is for because it is a header only send and reply.

QuantumEntangledAndy:
@thirtythreeforty I think we are mislabeling the binary field.

Have a look at message id 45

- 45: `<OsdChannelName>` (write)

  - Client

    - Header

        |    magic     |  message id  | message length | encryption offset | Encryption flag | Unknown | message class | Binary Offset |
        |--------------|--------------|----------------|-------------------|-----------------|---------|---------------|---------------|
        | 0a bc de f0  | 00 00 00 2d  |  00 00 02 23   |    32 00 00 00    |       00        |   00    |     64 14     |  00 00 00 68  |

    - Body

    ```xml
    <?xml version="1.0" encoding="UTF-8" ?>
    <Extension version="1.1">
    <channelId>0</channelId>
    </Extension>
    ```

    - Binary

    ```xml
    <?xml version="1.0" encoding="UTF-8" ?>
    <body>
    <OsdChannelName version="1.1">
    <channelId>0</channelId>
    <name>Cammy02</name>
    <enable>0</enable>
    <topLeftX>65536</topLeftX>
    <topLeftY>65536</topLeftY>
    <enBgcolor>0</enBgcolor>
    <enWatermark>0</enWatermark>
    </OsdChannelName>
    <OsdDatetime version="1.1">
    <channelId>0</channelId>
    <enable>1</enable>
    <topLeftX>65537</topLeftX>
    <topLeftY>1</topLeftY>
    <language>Chinese</language>
    </OsdDatetime>
    </body>
    ```

It contains both `body` and `binary` parts before and after the `binary_offset`. The `binary` is more like the `payload` and the `body` is more like the `meta` for the body which channel or which user etc.

Should we invest in the refractor required to the to generalise the two parts as XMLs and binaries. Msg ID 3 seems to make a bit more sense now as to why the binary is stateful based on the `meta` of the extension xml `<binary>1</binary>`. I think that the first msg block pre binary_offset is supposed to describe the payload to follow.

thirtythreeforty:
This is a very nice refactor, thank you. A couple questions:

- Would you mind posting your script? You don't have to commit it, just attached in a comment.
- Are you ready to merge this?
- @kickbits I didn't notice when you showed up. Are you @twistedddx ?

And a comment:

> As long as I stick to roughly the right messages I should be ok, just should avoid things like incorrect message lengths I don't know what sort of protections it has against buffer overflow etc.

If Reolink starts giving us grief in future releases, locking stuff down etc, I am pretty confident we can fuzz their Baichuan parser and find some interesting behavior ;)

twistedddx:
Yes kickbits is me.
I had issues with maintaining a fork of thirtythreeforty neolink and quantumentangledandy neolink. So I created a second account that could fork andy's.

QuantumEntangledAndy:
> - Would you mind posting your script? You don't have to commit it, just attached in a comment.


The script required changes to dissector so that it prints the relevant parts to stdout (as yml) which is then redirected to file and then my script just sorts and prints it.

I could post the script but without the relevant parts in the dissector it is mostly useless. I'm not happy with putting the parts into the dissector as making it print to stdout for everyone seems to be bad form. Maybe if I make it print to a wireshark tap instead that would be better.

QuantumEntangledAndy:
The dumping to yml and then parsing the yml was done with.

These changes to the dissector

<details>
  <summary>Changes to process_body and new function get_body</summary>

```lua
function get_body(header, body_buffer)
  if header.msg_len == 0 then
    return {
      msg = nil,
      binary = nil,
    }
  end

  local msg_data = nil

  local msg_len = header.bin_offset
  if msg_len == nil then
    msg_len = header.msg_len
  end
  local msg_buffer = body_buffer(0, msg_len)
  local msg_is_xml = false
  if msg_len > 0 then
    if msg_len >= 4 then
      if xml_encrypt(msg_buffer(0,5):bytes(), header.enc_offset):raw() == "<?xml" then -- Encrypted xml found
        msg_data = xml_encrypt(msg_buffer:bytes(), header.enc_offset)
        msg_is_xml = true
      elseif msg_buffer(0,5):string() == "<?xml" then  -- Unencrypted xml
        msg_data = msg_buffer
        msg_is_xml = true
      else
        msg_data = msg_buffer
      end
    else
      msg_data = msg_buffer
    end
  end

  local binary_buffer = nil
  local binary_is_xml = false
  local binary_data = nil
  if header.bin_offset ~= nil then
    local bin_len = header.msg_len - header.bin_offset
    if bin_len > 0 then
      binary_buffer = body_buffer(header.bin_offset, bin_len) -- Don't extend beyond msg size
      if bin_len > 4 then
        if xml_encrypt(binary_buffer(0,5):bytes(), header.enc_offset):raw() == "<?xml" then -- Encrypted xml found
          binary_data = xml_encrypt(binary_buffer:bytes(), header.enc_offset)
          binary_is_xml = true
        elseif binary_buffer(0,5):string() == "<?xml" then  -- Unencrypted xml
          binary_data = binary_buffer
          binary_is_xml = true
        else
          binary_data = binary_buffer
        end
      else
        binary_data = binary_buffer
      end
    end
  end
  local msg = nil
  if msg_data ~= nil then
    msg = {
      data = msg_data,
      buffer = msg_buffer,
      is_xml = msg_is_xml,
    }
  end
  local binary = nil
  if binary_data ~= nil then
    binary = {
      data = binary_data,
      buffer = binary_buffer,
      is_xml = binary_is_xml,
    }
  end
  return {
    msg = msg,
    binary = binary,
  }
end

function process_body(header, body_buffer, bc_subtree, pinfo)
  if header.msg_len == 0 then
    return
  end

  local body = bc_subtree:add(bc_protocol, body_buffer(0,header.msg_len),
    "Baichuan Message Body, " .. header.class .. ", length: " .. header.msg_len .. ", type " .. header.msg_type .. ", encrypted: " .. tostring(header.encrypted))

  local body_data = get_body(header, body_buffer)


  if header.class == "legacy" then
    if header.msg_type == 1 then
      body:add_le(username, body_buffer(0, 32))
      body:add_le(password, body_buffer(0 + 32, 32))
    end
  else
    if body_data.msg ~= nil then
      local msg_buffer = body_data.msg.buffer
      local body_tvb = msg_buffer:tvb("Payload")
      if body_data.msg.is_xml then
        body:add(body_tvb(), "XML")
        Dissector.get("xml"):call(body_data.msg.data:tvb("XML"), pinfo, body)
      else
        body:add(body_tvb(), "Binary")
      end
    end

    if body_data.binary ~= nil then
      local binary_buffer = body_data.binary.buffer
      local binary_tvb = binary_buffer:tvb("Binary Payload");
      if body_data.binary.is_xml then
        Dissector.get("xml"):call(body_data.binary.data:tvb("XML"), pinfo, body)
        body:add(binary_tvb(), "XML (in binary block)")
      else
        body:add(binary_tvb(), "Binary")
      end
    end
  end
end
```

</details>

With this at the end of the packet parsing to do the printing

<details>
  <summary>Printing part is quite hairy</summary>

```lua
-- Remaning bytes?
      if remaining == 0 then
        continue_loop = false
      else
        pinfo.cols['info'] = "Another packet"
        sub_buffer = sub_buffer(full_body_len, nil)
      end

      -- Print as yml for later parsing
      -- This print the first client and first server message of each id
      -- You will need to have wireshark writing to stdout (start it from the terminal)
      local should_print = false
      if pinfo.dst_port == 9000 then
        if client_seen[header.msg_type] == nil then
          should_print = true
          client_seen[header.msg_type] = true
        end
      else
        if server_seen[header.msg_type] == nil then
          should_print = true
          server_seen[header.msg_type] = true
        end
      end
      if should_print then
        print("")
        if pinfo.dst_port == 9000 then
          print("- client: true")
        else
          print("- client: false")
        end
        print("  magic: " .. header.magic)
        print("  msg_id: " .. header.msg_type)
        print("  msg_len:  " .. header.msg_len)
        print("  enc_offset:  " .. header.enc_offset)
        print("  enc_flag:  " .. header.enc_flag)
        print("  unknown:  " .. header.unknown)
        print("  msg_cls:  " .. header.msg_cls)
        if header.bin_offset ~= nil then
          print("  bin_offset:  " .. header.bin_offset)
        end
        if body_buffer ~= nil then
          local body = get_body(header, body_buffer)
          if body.msg ~= nil then
            if body.msg.is_xml then
              print("  msg:  |\n    " .. string.gsub(body.msg.data:raw(),"\n", "\n    "))
            else
              print("  msg:  |\n    " .. string.gsub(body.msg.data:bytes():tohex(),"\n", "\n    "))
            end
          end
          if body.binary ~= nil then
            if body.binary.is_xml then
              print("  binary:  |\n    " .. string.gsub(body.binary.data:raw(),"\n", "\n    "))
            else
              print("  binary:  |\n    " .. string.gsub(body.binary.data:bytes():tohex(),"\n", "\n    "))
            end
          end
        end
      end
    else
      pinfo.cols['info'] = "Invalid magic"
      return -- Not a valid header
    end
  end

  local msg_type_strs = table.concat(table_msg_type_str, ",")
  local msg_types = table.concat(table_msg_type, ",")
  pinfo.cols['info'] = msg_type_strs .. ", type " .. msg_types
end
```

</details>

This parsing script

<details>
  <summary>Its written in python just coz I am familiar with it</summary>

```python
#! /usr/bin/env python3
"""Just a quick yml parser for the dissectors dump."""

from pathlib import Path
from yaml import safe_load as load

path = Path("./wiredump.yml")
data = load(path.read_text())

data = sorted(data, key=lambda x: x['msg_id'])

known_names = {
    1: "login",
    2: "logout",
    3: "<Preview> (video)",
    4: "<Preview> (stop)",
    5: "<FileInfoList> (replay)",
    7: "<FileInfoList> (stop)",
    8: "<FileInfoList> (DL Video)",
    10: "<TalkAbility>",
    13: "<FileInfoList> (download)",
    14: "<FileInfoList>",
    15: "<FileInfoList>",
    16: "<FileInfoList>",
    25: "<VideoInput> (write)",
    26: "<VideoInput>",
    31: "Start Motion Alarm",
    33: "<AlarmEventList>",
    36: "<ServerPort> (write)",
    37: "<ServerPort>",
    38: "<Ntp>",
    39: "<Ntp> (write)",
    40: "<Ddns>",
    41: "<Ddns> (write)",
    42: "<Email>",
    43: "<Email> (write)",
    44: "<OsdChannelName>",
    45: "<OsdChannelName> (write)",
    46: "<MD>",
    47: "<MD> (write)",
    50: "<VideoLoss>",
    51: "<VideoLoss> (write)",
    52: "<Shelter>",
    54: "<RecordCfg>",
    55: "<RecordCfg> (write)",
    56: "<Compression>",
    57: "<Compression> (write)",
    58: "<AbilitySupport>",
    59: "<UserList> (write)",
    65: "<ConfigFileInfo> (Export)",
    66: "<ConfigFileInfo> (Import)",
    68: "<Ftp>",
    69: "<Ftp> (write)",
    70: "<FtpTask>",
    71: "<FtpTask> (write)",
    76: "<Ip>",
    77: "<Ip> (write)",
    78: "<VideoInput>",
    79: "<Serial> (ptz)",
    80: "<VersionInfo>",
    81: "<Record> (schedule)",
    82: "<Record> (write)",
    83: "<HandleException>",
    84: "<HandleException> (write)",
    91: "<DisplayOutput>",
    92: "<DisplayOutput> (write)",
    93: "<LinkType>",
    97: "<Upnp>",
    98: "<Upnp> (write)",
    100: "<AutoReboot> (write)",
    101: "<AutoReboot>",
    102: "<HDDInfoList>",
    104: "<SystemGeneral>",
    105: "<SystemGeneral> (write)",
    106: "<Dst>",
    107: "<Dst> (write)",
    108: "<ConfigFileInfo>",
    114: "<Uid>",
    115: "<WifiSignal>",
    120: "<OnlineUserList>",
    122: "<PerformanceInfo>",
    123: "<ReplaySeek>",
    132: "<VideoInput>",
    133: "<RfAlarm>",
    141: "<Email> (test)",
    142: "<DayRecords>",
    145: "<ChannelInfoList>",
    146: "<StreamInfoList>",
    151: "<AbilityInfo>",
    190: "PTZ Preset",
    194: "<Ftp> (test)",
    199: "<Support>",
    208: "<LedState>",
    210: "<PTOP>",
    211: "<PTOP> (write)",
    219: "<PushTask>",
    228: "<Crop>",
    229: "<Crop> (write)",
    230: "<cropSnap>",
    272: "<findAlarmVideo>",
    273: "<alarmVideoInfo>",
    274: "<findAlarmVideo>",
}


def chunk(lst, n):
    """Chunks the list."""
    n = max(1, n)
    return (lst[i:i+n] for i in range(0, len(lst), n))


def hex_fmt(num, expected):
    """Format the number as hex."""
    hexy = hex(num).replace('0x', '')
    if len(hexy) % 2 != 0:
        hexy = f"0{hexy}"
    if len(hexy)//2 != expected:
        for i in range(expected-len(hexy)//2):
            hexy = f"00{hexy}"
    return ' '.join(chunk(hexy, 2))


prev_num = -1
for d in data:
    magic = hex_fmt(d['magic'], 4)
    msg_id_num = d['msg_id']
    msg_id = hex_fmt(d['msg_id'], 4)
    msg_len = hex_fmt(d['msg_len'], 4)
    enc_off = hex_fmt(d['enc_offset'], 4)
    enc_flag = hex_fmt(d['enc_flag'], 1)
    unknown = hex_fmt(d['unknown'], 1)
    msg_class = hex_fmt(d['msg_cls'], 2)

    if prev_num != msg_id_num:
        known_name = known_names.get(msg_id_num, '')
        print(f"\n- {msg_id_num}: {known_name}")
    prev_num = msg_id_num
    if d['client']:
        print("\n  - Client")
    else:
        print("\n  - Camera")
    if 'bin_offset' not in d:
        print(f"""\n    - Header

    |    magic     |  message id  | message length | encryption offset | Encryption flag | Unknown | message class |
    |--------------|--------------|----------------|-------------------|-----------------|---------|---------------|
    | {magic:11}  | {msg_id:11}  |  {msg_len:11}   |    {enc_off:11}    |       {enc_flag:2}        |   {unknown:2}    |     {msg_class:5}     |""") # noqa
    else:
        bin_off = hex_fmt(d['bin_offset'], 4)
        print(f"""\n    - Header

        |    magic     |  message id  | message length | encryption offset | Encryption flag | Unknown | message class | Binary Offset |
        |--------------|--------------|----------------|-------------------|-----------------|---------|---------------|---------------|
        | {magic:11}  | {msg_id:11}  |  {msg_len:11}   |    {enc_off:11}    |       {enc_flag:2}        |   {unknown:2}    |     {msg_class:5}     |  {bin_off:11}  |""") # noqa

    if 'msg' in d:
        print("\n    - Body")
        body = d['msg']
        is_xml = body[0:5] == "<?xml"
        if is_xml:
            padded_body = '\n'.join(
                [f"    {line}" for line in body.strip().split('\n')]
            )
            print("\n    ```xml")
            print(padded_body)
            print("    ```")
        else:
            padded_body = '\n'.join(
                [f"    {line}" for line in chunk(body.strip(), 64)]
            )
            print("\n    ```hex")
            print(padded_body)
            print("    ```")

    if 'binary' in d:
        print("\n    - Binary")
        body = d['binary']
        is_xml = body[0:5] == "<?xml"
        if is_xml:
            padded_body = '\n'.join(
                [f"    {line}" for line in body.strip().split('\n')]
            )
            print("\n    ```xml")
            print(padded_body)
            print("    ```")
        else:
            padded_body = '\n'.join(
                [f"    {line}" for line in chunk(body.strip(), 64)]
            )
            print("\n    ```hex")
            print(padded_body)
            print("    ```")
```

</details>

As I said not pretty haha

QuantumEntangledAndy:
From our discussion and investigation in #78 I have added a few more fields to the header

twistedddx:
I think the BC protocol discussion belongs here rather than adding noise to the MQTT PR.

You asked why BC would want to waste a byte in every packet for what stream type is currently playing.
I imagine it would be useful to split up the data if you had HD and SD streams both going at the same time.

1 byte which camera
1 byte which stream from that camera
1 byte unknown
1 byte message handle (client increments almost every time the client sends a message)

Every camera is a single channel NVR.
That NVR has a single camera on channel_id 0 which is itself.
So the 1st byte would always be 00 when direct connecting to a camera as channel_id 0 is the only channel it has.

For NVR that byte is set in both directions of communication.
Eg Client sends OsdChannelName to channel_id 03
03 00 00 1d
NVR responds (on behalf of camera)
03 00 00 1d


For the "encrypted" byte I believe:
00 = XML in payload 1
c8 = XML in payload 2

Excluding login which may actually be encrypted related.
Eg
Client sends 01 for desire to do encryption.
Device responds either 01 for encrypted, 00 for not encrypted.
And then the Client sends 00 and XML in payload 2 to be the odd one out during modern login.

c8 = 200 and may be related to BC version 2.0.0 which came out in about 2016 and may have been an evolution of BC protocol.

twistedddx:
Tell me why it appears BC write the NVR channel count into every P frame header.
Why would it be important for the video to know it was connected to an 8 channel recorder?
And the camera itself is putting that in to the video stream, it isn't the NVR writing that in to the video stream.

Camera connects to NVR and learns its NVR is 8 channels.
Now the Camera will write 08 at byte 13 in the BC media P frame header on any stream it sends.

twistedddx:
Actually scrap the 00/c8 bit I believe this is better:
First 2 messages in login appear to be encrypted related.
After that client always uses 00 and camera always uses c8 00 for OK and 90 01 for NOT OK
That is:
200 OK
400 Bad Request

QuantumEntangledAndy:
This is indeed the better place I think too. When I get the time I'll document the changes to the payload layout here too

> You asked why BC would want to waste a byte in every packet for what stream type is currently playing.

With regards to this I querying it because it seems useless to add stream quality into non stream related messages. Why include it into the WiFi strength or PTZ messages. It is also redundant, if the client requested SD with message handle 10 then all messages handle 10 from that channel should be SD no need to encode in header. Anyways it's probably redundancy to ensure the client doesn't randomly encode a large h264 mid stream and ruin the video stream.

Anyways I digress...

Back to the more fun of the new message handle and friends and looking forward to trying to use this to get both streams at the same time from the camera. 

With regards to the encrupted/success byte I will look into how it works on my E1 too when I get the time. I wonder if the camera will send 00 first in pakcets that the camera sends but the client didn't request or weather it's always client 00. I think the camera sends some stream info right after login that the client doesn't request so I should look into those messages. 

Current theory is:

For 20 length header this byte is encryption request (login header)

For 24 byte header (all other message)

- 00 Publish Request

- c8 OK

- 90 Bad Request

QuantumEntangledAndy:
During the legacy login on the encryption/status byte, my E1 sends 01dc and the camera replies with 01dd do you observe the same @twistedddx? What is the reply when threre is no encryptio

During modern modern my E1 always send 00 and the camera always replies c800 so in modern it seems to always be HTML status codes

twistedddx:
rlc-520/rlc-410
01dc
01dd

NHD-818
01dc
00dd

QuantumEntangledAndy:
So always dc and dd, I wonder if this dd and dc makes more sense in contex of the other legacy messages but probable beyond scope and the moment for our purposes it seems this part is indeed the encryption flag and its only send during login

thirtythreeforty:
> > * Would you mind posting your script? You don't have to commit it, just attached in a comment.
> 
> The script required changes to dissector so that it prints the relevant parts to stdout (as yml) which is then redirected to file and then my script just sorts and prints it.
> 
> I could post the script but without the relevant parts in the dissector it is mostly useless. I'm not happy with putting the parts into the dissector as making it print to stdout for everyone seems to be bad form. Maybe if I make it print to a wireshark tap instead that would be better.

Sounds good, no worries.

thirtythreeforty:
@QuantumEntangledAndy I'm happy with what you have; are you done making changes to this PR (you can always make another ;) or would you like me to wait to merge this?

thirtythreeforty:
> After that client always uses 00 00 and camera always uses `c8 00` for OK and `90 01` for NOT OK
> That is:
> 200 OK
> 400 Bad Request

Yes! This is definitely the "response code" and they are using HTTP-ish result codes. I've never seen anything other than 0 = request, 200 = Ok, 400 = Error.

thirtythreeforty (approved):

thirtythreeforty:
I would prefer to squash this PR unless you think otherwise.

twistedddx:
I checked a legacy protocol camera (nhd-805) and the 01dc makes no more sense there. 
I see
Client login 01dc. 
Camera login response is c800, this is the packet that is just a block of text with C0C0C0C0 in "bin offset" area. All headers look to be 20 byte in fact, the C0C0C0C0 is just part of its payload. 
Client response to cameras text is 9001. First time I have ever seen the client not use 0000.

Client and camera continue with 0000 c800 etc. There is no 00dd type response etc. 

I have no interest in reversing the legacy protocol and it is not giving new insight to the modern protocol in this case.

QuantumEntangledAndy:
Squashing is fine by me no need to include so many commits for what is mostly a doc update. @twistedddx just added another commit to rename the encryption flag to status code in modern messages. If that's thats the last one he has coming please merge and we can open a new PR with more updates.

I am thinking to make another PR soon that reworks these changes into the neolink code. Let us access the status codes and the msg handle for the replies. Also rework binary to to main payload and what's currently called XML to extension-meta.

twistedddx:
That is my last updates.

The things you find when you go digging amazes me. 
Bc stuffed up their xmlencryptionoffset. Surely they meant to change the offset per message handle and instead did it per channel. A standalone camera only uses channel 0 so the offset does nothing in most cases.

QuantumEntangledAndy:
Yes it's quite poor and message id would make more sense. However the encryption only uses the first 4 bits of the byte in the offset so it would still only have 16 different encryptions.

I think I remember you mentioning that there were only 8 channels in an NVR? Maybe they wanted to use a variable that was limited to 8 to avoid doing the bitmask for the first 4 bits.

twistedddx:
My Nvr is 8 channels but BC have had 4, 8 and 16 channels for quite some time.

QuantumEntangledAndy:
16 channels is a 4 but uint just like the encryption offset field. Still perhaps unrelated we will never really know why theyre using the channel for encryption

PR #84, opened 2020-10-16
Make Dockerfile compatible with ARM

### LINKED ISSUE #73: Raspberry Pi support for the docker image

JoaoNogueira114:
Docker image docker.io/rust:1-alpine was not compatible with ARM.
Using the base image docker.io/alpine and installing rust instead will allow the image to be build for both ARM and x86.

Tested and working on a Pi model 3B, running Raspbian GNU/Linux 9.13

Closes #73

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,9 +2,10 @@
 # Copyright (c) 2020 George Hilliard
 # SPDX-License-Identifier: AGPL-3.0-only
 
-FROM docker.io/rust:1-alpine AS build
+FROM docker.io/alpine AS build
 MAINTAINER thirtythreeforty@gmail.com
 
+RUN apk add rust cargo
 RUN apk add --no-cache -X http://dl-cdn.alpinelinux.org/alpine/edge/testing \
   gst-rtsp-server-dev
 RUN apk add --no-cache musl-dev gcc
@@ -34,7 +35,9 @@ COPY --from=build \
   /usr/local/src/neolink/target/release/neolink \
   /usr/local/bin/neolink
 COPY docker/entrypoint.sh /entrypoint.sh
+RUN chmod +x /entrypoint.sh
 
 CMD ["/usr/local/bin/neolink", "--config", "/etc/neolink.toml"]
 ENTRYPOINT ["/entrypoint.sh"]
-EXPOSE 8554 
+EXPOSE 8554
+


QuantumEntangledAndy:
Can I ask how you build this for arm within github actions without using a raspberry pi? Perhaps I'm missing something.

QuantumEntangledAndy:
I think youd need to also make the GitHub action part use qemu and xbuild too. Could you have a look at the changes I tried in #76 and add those to the docker.yml in GitHub actions. Your best was to test is to use GitHub actions in your own repo.

QuantumEntangledAndy:
Something like this

```yml
- name: Set up QEMU
  if: ${{ steps.vars.outputs.HAS_SECRET_TOKEN }}
  uses: docker/setup-qemu-action@v1
- name: Set up Docker Buildx
  if: ${{ steps.vars.outputs.HAS_SECRET_TOKEN }}
  uses: docker/setup-buildx-action@v1
- name: Login to dockerhub
  if: ${{ steps.vars.outputs.HAS_SECRET_TOKEN }}
  uses: docker/login-action@v1
  with:
    username: ${{ steps.string_user.outputs.lowercase }}
    password: ${{ secrets.DOCKER_TOKEN }}
- name: Get tag name
  if: ${{ steps.vars.outputs.HAS_SECRET_TOKEN }}
  id: tags
  run: |
    branch="${GITHUB_REF#refs/heads/}"
    tags="${REPO_NAME}:${branch}"
    if [ "${branch}" == "master" ]; then
      tags="${tags},${REPO_NAME}:latest"
    fi
    echo "::set-output name=TAGS::${tags}"
  env:
    REPO_NAME: ${{ steps.string_repo.outputs.lowercase }}
- name: Push to Docker Hub Debian
  if: ${{ steps.vars.outputs.HAS_SECRET_TOKEN }}
  uses: docker/build-push-action@v2
  with:
    platforms: linux/arm/v7,linux/amd64
    push: true
    file: docker/debian/Dockerfile
    tags: ${{ steps.tags.outputs.TAGS }}
```

So that the build machine makes both images before publishing to dockerhub

QuantumEntangledAndy:
I would be interested in seeing if the alpine also runs into the qemu issues I was having #76 with the cargo build step failing due to inode sizes

QuantumEntangledAndy:
Could you try with armv6 too. I had already gotten amd64/armv7/armv8 working. It was just armv6 for someone on a raspberry pi 1 that was causing me trouble.

QuantumEntangledAndy:
I think I can see from your commit that you are trying to split armv6 out as its own build. When I tried something similar it overwrote the build on dockerhub rather than added to it. I'm sure there's a way to do it but I didn't look into it. Please tell me if you can manage this as I'd be interested.

Also are you breaking it out as its own build due to the build time? I found armv6 took over 7 hours. If we can get it to publish incrementally to docker hub maybe using a matrix would be more efficient than I seperate job.

JoaoNogueira114:
Yes, that is what I was trying to do. When I bundled all builds together it timed out at 6h, however when I split them the times are:
- arm v6: 2h
- arm v7, v8 and amd64: 3h45m
but the second build overwrote the first, as you reported.

I had never used github actions before so I am not sure yet what is causing this issues. But I will look into it in the next couple of days.

QuantumEntangledAndy:
I think the overriding is part of docker rather than GitHub. I suspect that there is an option in docker to tell it to add to a published build but I haven't looked into it either yet.

QuantumEntangledAndy:
This [link](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/) seems to suggest that we build to different tags then crate a manifest that points to this different tags

QuantumEntangledAndy:
Rather than making the manifest as described in that link we could go down the route of build on docker in GitHub job 1 and save the binary as an artifact. Then in job 2 cp in the binary to the docker and publish.

QuantumEntangledAndy:
Personally though I think this is more effort then it is worth. Let's just leave it as amd64 and armv7 and add others at a later date.

JoaoNogueira114:
I may take a final look during the weekend but ultimately I agree with you. Releasing it for amd64/armv7/armv8 will probably be enough for most users.

QuantumEntangledAndy:
I've just noticed your docker-arm-v6.yaml the way this is setup won't it ovewrite the amd64/armhf/arm64 image on dockerhub when it gets pushed. I think perhaps we should remove that before merging this

PR #87, opened 2020-10-29
More detailed error messages on deser fail and timeout

QuantumEntangledAndy:
While I was making changes to the deserialiser I noticed that any errors during derserialisation would be silently ignored. Instead it would just drop the connection and restart.

This makes changes so that it prints the error prior to dropping and restarting the connection.

I also noticed that the error `std::sync::mpsc::RecvTimeoutError` is actually an enum of, `Timeout` or `Disconnected` but in the code we treat them the same as `Timeout` (this is why sometimes it would say timeout error even when the the 5 seconds (of `RX_TIMEOUT`) had yet to pass)

This also makes changes to the `Error` enum so that we correctly distinguish between the two.

### DIFF
--- a/src/bc_protocol/connection.rs
+++ b/src/bc_protocol/connection.rs
@@ -55,7 +55,15 @@ impl BcConnection {
 
         let rx_thread = std::thread::spawn(move || {
             let mut context = BcContext::new();
-            while BcConnection::poll(&mut context, &conn, &mut subs).is_ok() {}
+            while {
+                match BcConnection::poll(&mut context, &conn, &mut subs) {
+                    Ok(_) => true,
+                    Err(err) => {
+                        error!("Deserialization error: {}", err);
+                        false
+                    }
+                }
+            } {}
         });
 
         Ok(BcConnection {
--- a/src/bc_protocol/connection.rs
+++ b/src/bc_protocol/connection.rs
@@ -59,7 +59,7 @@ impl BcConnection {
                 match BcConnection::poll(&mut context, &conn, &mut subs) {
                     Ok(_) => true,
                     Err(err) => {
-                        error!("Deserialization error: {}", err);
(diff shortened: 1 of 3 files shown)


src/bc_protocol/connection.rs
```diff
@@ -55,7 +55,15 @@ impl BcConnection {
...
-            while BcConnection::poll(&mut context, &conn, &mut subs).is_ok() {}
+            while {
+                match BcConnection::poll(&mut context, &conn, &mut subs) {
+                    Ok(_) => true,
+                    Err(err) => {
+                        error!("Deserialization error: {:?}", err);
+                        false
+                    }
+                }
+            } {}
```
thirtythreeforty, line 66: I see what you're doing here, but the "naked" `while` is weird imho.  How about:
```
    let mut result;
    while { result = foo(); result.is_ok() } {}
    error!("Deserialization error: {:?}", result.unwrap_err);
    result.unwrap_err();
```

thirtythreeforty (changes requested):
Small quibble, see above :)

src/bc_protocol/connection.rs
QuantumEntangledAndy: I like being verbose but alright I'll change it haha

QuantumEntangledAndy:
Edited to less naked loop

src/bc_protocol/connection.rs
QuantumEntangledAndy: Done

QuantumEntangledAndy:
I would also like to get a full backtrace with the error. Would you be OK with swapping from `err-derive` to something that supports backtraces?

QuantumEntangledAndy:
Ugh backtraces are nightly only... why rust why...

thirtythreeforty (approved):

PR #93, opened 2020-11-25
Dissector - Add BC to BC magic 0x0fedcba0

twistedddx:
Connection between NVR and IPC on NVR LAN use this magic instead of 0x0abcdef0

When NVR does login to IPC, NVR has encrypt flag set to false and camera responds with encrypt set to false.
All XML continues to be plain text.

This was tested on RLN8-410 running Oct 2020 firmware and RLC-410 5mp running October 2020 firmware.


thirtythreeforty (approved):
How interesting. I wonder if there's anything the NVR can do that the client can't.

PR #94, opened 2020-11-25
Add UDP dissector

QuantumEntangledAndy:
This add a UDP dissector. It is not perfect (because packets have to be reassembled manually for UDP) but it should suffice.

We can thank  jdillenkofer/camera_proxy for his work on UDP, which he kindly allowed us to use.

[jdillenkofer/camera_proxy](github.com/jdillenkofer/camera_proxy/blob/af9a170d3a8db831318a1aafcb2bf097ab62fd40/wireshark-dissector.lua)

### DIFF
--- a/dissector/baichuan.lua
+++ b/dissector/baichuan.lua
@@ -19,6 +19,18 @@ f_bin_offset = ProtoField.int32("baichuan.bin_offset", "binOffset", base.DEC)
 username = ProtoField.string("baichuan.username", "username", base.ASCII)
 password = ProtoField.string("baichuan.password", "password", base.ASCII)
 
+-- UDP Related content
+udp_magic = ProtoField.int32("baichuan.udp_magic", "udp_magic", base.DEC)
+udp_type = ProtoField.int8("baichuan.udp_type", "udp_type", base.DEC)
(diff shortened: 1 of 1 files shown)


awlx:
As I just got a camera with UDP, how do I try this? :)

QuantumEntangledAndy:
@awlx This is a PR, as in a pull request, to add a dissector to wireshark. So your comment is in the wrong place. 

We don't currently support UDP for neolink. It's on our Todo list and we understand the protocol but it's not being actively worked on at the moment. I'd like to tackle it but I'm too busy with work right now and I don't even own a UDP camera to test so it will be slow going.

awlx:
I am sorry, was just searching through all the PRs and code and didn't really take a too deep look ;). If you send me a message to aw @ awlnx.space we can maybe arrange shipping you one of the cameras :).

QuantumEntangledAndy:
Thanks for the offer but I'm not too comfortable with accepting. It would kinda feel like I'd have to it asap and probably more expensive then it should be as I live in Thailand. Also thirtythreeforty has one already so we can test it on that once we have gotten the code in. Thank you though.

PR #99, opened 2020-12-03
config: allow hostnames for camera addresses

kevans91:
This offers more flexibility, as a camera system admin only
needs a DNS assignment rather than DHCP. The syntax is still
as expected:

address = "cameras.example.com:9000"

This is especially handy in more complex topologies, where one
or more specified "camera" may actually be pointed at a system
providing a tunnel through to multiple cameras; names make it
easier to identify said gateway machine.

All IPs that the given name resolves to are tried so that we
correctly handle, e.g., a name with both AAAA and A records.

Note that addresses are no longer validated up front, but instead
simply get resolved when the connection attempt is made.

Name resolution will be aborted after ~60s of failure.

Note that my camera setup is not yet complete, so this was only tested against:
- Non-existent IP
- Unresolvable hostname
- localhost w/ port closed
- localhost w/ port open

All of these behave as expected. Please excuse my rust, as I'm quite new at it.

### DIFF
--- a/src/bc_protocol.rs
+++ b/src/bc_protocol.rs
@@ -58,6 +58,9 @@ pub enum Error {
     #[error(display = "Credential error")]
     AuthFailed,
 
+    #[error(display = "Failed to translate camera address")]
+    AddrResolutionError,
+
     #[error(display = "ADPCM Decoding Error")]
     AdpcmDecodingError(&'static str),
 
@@ -81,22 +84,34 @@ impl Drop for BcCamera {
 }
 
 impl BcCamera {
-    pub fn new_with_addr<T: ToSocketAddrs>(hostname: T) -> Result<Self> {
-        let address = hostname
-            .to_socket_addrs()?
-            .next()
-            .ok_or(Error::Other("Address resolution failed"))?;
+    pub fn connect<T: ToSocketAddrs>(host: T) -> Result<BcCamera> {
+        let addr_iter = match host.to_socket_addrs() {
+            Ok(iter) => iter,
+            Err(_) => return Err(Error::AddrResolutionError),
+        };
 
-        Ok(Self {
-            address,
-            connection: None,
-            logged_in: false,
-        })
-    }
+        for addr in addr_iter {
+            debug!("Trying {}", addr);
+            let conn = match BcConnection::new(addr, RX_TIMEOUT) {
+                Ok(conn) => conn,
+                Err(err) => match err {
+                    connection::Error::CommunicationError(ref err) => {
+                        debug!("Assuming timeout from {}", err);
+                        continue;
+                    }
+                    err => return Err(err.into()),
+                },
+            };
 
-    pub fn connect(&mut self) -> Result<()> {
-        self.connection = Some(BcConnection::new(self.address, RX_TIMEOUT)?);
-        Ok(())
+            debug!("Success: {}", addr);
+            return Ok(Self {
+                address: addr,
+                connection: Some(conn),
+                logged_in: false,
+            });
+        }
+
+        return Err(Error::Timeout);
     }
 
     pub fn disconnect(&mut self) {
--- a/src/config.rs
+++ b/src/config.rs
@@ -2,7 +2,6 @@ use lazy_static::lazy_static;
 use regex::Regex;
 use serde::Deserialize;
 use std::clone::Clone;
-use std::net::SocketAddr;
 use std::time::Duration;
 use validator::{Validate, ValidationError};
 use validator_derive::Validate;
@@ -45,7 +44,7 @@ pub struct CameraConfig {
     pub name: String,
 
     #[serde(rename = "address")]
-    pub camera_addr: SocketAddr,
+    pub camera_addr: String,
 
     pub username: String,
     pub password: Option<String>,
--- a/src/main.rs
+++ b/src/main.rs
@@ -198,7 +198,6 @@ fn camera_main(
 ) -> Result<Never, CameraErr> {
     let mut connected = false;
     (|| {
-        let mut camera = BcCamera::new_with_addr(camera_config.camera_addr)?;
         if camera_config.timeout.is_some() {
             warn!("The undocumented `timeout` config option has been removed and is no longer needed.");
             warn!("Please update your config file.");
@@ -208,7 +207,8 @@ fn camera_main(
             "{}: Connecting to camera at {}",
             camera_config.name, camera_config.camera_addr
         );
-        camera.connect()?;
+
+        let mut camera = BcCamera::connect(&camera_config.camera_addr)?;
 
         camera.login(&camera_config.username, camera_config.password.as_deref())?;
 


src/main.rs
```diff
@@ -127,12 +129,46 @@ fn camera_loop(
...
                 return Err(cam_err.err.into());
             }
-            _ => error!(
-                "Error streaming from camera {}, will retry in {}s: {}",
-                camera_config.name,
-                current_backoff.as_secs(),
-                cam_err.err
-            ),
+            neolink::Error::AddrResolutionError => {
+                // Special handling for halting resolution after the resolve timeout period elapses.
```
thirtythreeforty, line 133: I disagree with the strategy of "stop trying to resolve after a while."  There are many transient conditions that could cause a name resolution failure (DNS server hosed, unpropagated, etc), and we'd like to retry after those. Is there any reason this couldn't simply be removed?

In fact, I'm on the fence about the AuthFailed bailout, for similar reasons.

src/config.rs
```diff
@@ -45,7 +44,7 @@ pub struct CameraConfig {
     pub name: String,
 
     #[serde(rename = "address")]
-    pub camera_addr: SocketAddr,
+    pub camera_addr: String,
```
thirtythreeforty, line 47: This change removes all validation from parsing the camera address. (That is, if `SocketAddr` failed to parse, you'd get a "malformed config file" message, which I like.)  Question: Is there any validation that can be done, or is any string potentially a valid address?

src/bc_protocol.rs
```diff
@@ -74,29 +78,48 @@ impl<'a> From<std::sync::mpsc::RecvTimeoutError> for Error {
...
+                        continue;
+                    }
+                    err => return Err(err.into()),
+                },
+            };
+
+            debug!("Success: {}", addr);
+            self.address = Some(addr);
+            self.connection = Some(conn);
+            return Ok({});
```
thirtythreeforty, line 119: Nitpick: The idiom is `Ok(())` ("return an `Ok` containing `()`").

src/bc_protocol.rs
```diff
@@ -17,8 +17,9 @@ mod connection;
 mod media_packet;
 mod time;
 
-pub struct BcCamera {
-    address: SocketAddr,
+pub struct BcCamera<T: ToSocketAddrs> {
```
thirtythreeforty, line 20: I would prefer not to parameterize this struct. It's basically the runtime representation of the camera - and we don't have different kinds of cameras distinguished by the way they're named.

Instead, can you modify [`BcConnection::connect`](https://github.com/thirtythreeforty/neolink/blob/master/src/bc_protocol/connection.rs#L49) to accept a `ToSocketAddrs`? Then, you could pass the string straight in from the struct.

Since we create the socket with extra options, you'll still need to iterate over them in our connect_to helper. (This is [what `TcpStream::connect(ToSocketAddrs)` does](https://doc.rust-lang.org/src/std/net/tcp.rs.html#106), so that's good behavior.)

thirtythreeforty (changes requested):
Thanks for the PR! Accepting hostnames is a great suggestion. See my comments if you don't mind. (Also, if you don't have time to tackle more changes, let me know and I'm happy to run with what you have.)

src/bc_protocol.rs
thirtythreeforty: Oh, also you'll need to remove [my previous attempt](https://github.com/thirtythreeforty/neolink/blob/32400971a532509aade690b01c84c04c757dfc1c/src/bc_protocol.rs#L84) at what you're implementing. The correct behavior is indeed to iterate over the results of the name resolution.

src/config.rs
kevans91: Yeah, so the problem here is that it's actually really hard to pass an IP/name that isn't valid. e.g. observe `ping 6`, `ping 6.6`. Even worse, the POSIX-specified name resolution interface (getaddrinfo) doesn't require any particular format, it only notes that DNS *might* be the provider.

I don't have any non-FreeBSD systems around to see if it holds for GNU/Linux, but a silly example that's perfectly valid and usable here is:
```
# ping 'x!foo'
ping: Unknown host
# echo '127.0.0.1 x!foo' >> /etc/hosts
# ping 'x!foo'
PING x!foo (127.0.0.1): 56 data bytes
64 bytes from 127.0.0.1: icmp_seq=0 ttl=64 time=0.203 ms
64 bytes from 127.0.0.1: icmp_seq=1 ttl=64 time=117.458 ms
...
```
I had considered doing something like the following, but had somewhat reached my daily limit of Rust at the time :-) e.g. (prepare for an abomination of a mix of pseudocode and an attempt at rust):

```
if addr matches "^[-A-za-z0-9.]:[0-9]+$" || addr.parse::<Ipv4Addr>().is_ok() || addr.parse::<Ipv6Addr>().is_ok() {
    ok
}
Err(...)
```

This avoids the problem I had in my earlier prototypes where my custom validator just .to_socket_addrs() that sucker and couldn't discern potentially transient DNS errors vs. validation errors.

IMHO it's worth validating; your userbase is probably incredibly unlikely to abuse the system resolver like that, and there's a lot more value in calling it an invalid hostname for those that wouldn't. Those of us that would fiddle with it will immediately recognize the issue if the verbiage is correct, and those that wouldn't fiddle will quickly see that there's a typo.

kevans91:
Thanks for the review! I will implement the rest of your comments within a ~day or so. I have no compelling reason to timeout after a minute, so I will remove that.

src/config.rs
thirtythreeforty: This is a really good point. Had no idea you could put line noise in the hosts file (and have it work).

Regarding errors, I think either approach is okay as long as it loudly complains about unresolvable hostnames (in clear language - even "unresolvable hostname" is a bit obtuse). Since this is a runtime problem, and given your point above, I would lean toward accepting the config file and erroring on connect.

In other words, how you have it now :)

kevans91: I changed up the display string of the AddrResolutionError, but it might be too verbose: "Failed to translate camera address" -> I /think/ that strikes a decent balance between "what does that mean?" and "WHAT does that mean?" -- that might also be a step in the wrong direction, though. Thoughts?

src/bc_protocol.rs
kevans91: I switched it up to just providing a standard new() with some sensible default values, but then it occurred to me that you probably meant to just construct a new BcConnection if BcConnection::connect() succeeds since we will destroy and reconstruct a new BcConnection on failure anyways.

src/config.rs
thirtythreeforty: Yeah, it's tricky to come up with a good error message. I think that's a decent balance too. (I don't really have a better suggestion.) Looks good!

thirtythreeforty (approved):
Thanks very much! This looks great.

kevans91:
Excellent, many thanks! :-)

PR #103, opened 2020-12-09
Implement more CI builds: for macOS and ARM Linux

thirtythreeforty:
Rebases, extends and resolves #59. Thanks @QuantumEntangledAndy for getting this working!


QuantumEntangledAndy:
Looking good. I see what you mean by the matrix context now. I didn't realise you could put the variable in the key name too, so learnt a new thing :)

thirtythreeforty:
For some reason I can't quite get the i386-on-x64 build to replicate. You had it working and my refactor broke it... The underlying problem is that dpkg doesn't do what is asked of it.

.github/workflows/build.yaml
```diff
@@ -47,3 +55,65 @@ jobs:
...
+            target: ${{ matrix.target }}
+            profile: minimal
+      - name: Build
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --release --all-features --target=${{ matrix.target }}
+        env:
+          CC_${{ matrix.triple }}: ${{ matrix.gcc }}-gcc
+          CXX_${{ matrix.triple }}: ${{ matrix.gcc }}-g++
```
QuantumEntangledAndy, line 109: `matrix.triple` is undefined so this just becomes `CC_ = ...`

.github/workflows/build.yaml
```diff
@@ -47,3 +55,65 @@ jobs:
...
+          PKG_CONFIG_ALLOW_CROSS: 1
+          # Sometimes dpkg selects the native architecture despite our :arch
+          # qualifier if it thinks they're compatible (e.g. targetting x86 on
+          # x64 host). So also include the host's pkgconfig dir.
+          PKG_CONFIG_PATH: /usr/lib/${{ matrix.gcc }}/pkgconfig:/usr/share/pkgconfig
+          PKG_CONFIG_LIBDIR: /usr/lib/${{ matrix.gcc }}/pkgconfig:/usr/share/pkgconfig
```
QuantumEntangledAndy, line 115: pkg config path  is not the same as matrix.gcc.

matrix.gcc = i686-linux-gnu
correct = i386-linux-gnu

QuantumEntangledAndy (changes requested):

.github/workflows/build.yaml
thirtythreeforty: Ha well I guess these lines aren't needed! 'swhat I get for trying this at 11:30pm

.github/workflows/build.yaml
thirtythreeforty: Ok, see, this makes no sense to me - why shouldn't it match the compiler prefix? Maybe something to do with dpkg being too clever for its own good again...

Thanks. It would have taken me a while to suss this out.

.cargo/config
```diff
@@ -1,2 +1,12 @@
...
+
+# TODO not good as a default
+[target.armv7-unknown-linux-gnueabihf]
+linker = "arm-linux-gnueabihf-gcc"
+
+[target.aarch64-unknown-linux-gnu]
+linker = "aarch64-linux-gnu-gcc"
+
+[target.i686-unknown-linux-gnu]
+linker = "i686-linux-gnu-gcc"
```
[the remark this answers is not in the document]
thirtythreeforty: Unfortunately this doesn't work:
```
Error: .github/workflows/build.yaml (Line: 115, Col: 11): Unexpected value ''
Error: The template is not valid. .github/workflows/build.yaml (Line: 115, Col: 11): Unexpected value ''
```

Looks like the GitHub template expansion happens before any steps run.

thirtythreeforty:
Docker build is busted, but I'll open a new issue for that.

PR #107, opened 2020-12-14
Update documentation for status code and additional encryption offset data

twistedddx:


thirtythreeforty:
The documentation update is good, I will have to update the Baichuan parser based on the status fields you figured out. Thanks.

thirtythreeforty (approved):

twistedddx:
Dissector does already include status code from a previous commit. Although at the time I had called it "error_code".

thirtythreeforty:
True. I meant in Neolink itself.

QuantumEntangledAndy:
I have started adding status codes and message handle to neolink already as part as  #83 but never have had time to finish it :(. Buzy work things these days

Issue #37, opened 2020-07-12
RTSP TCP support


m1k1o:
Switching to TCP (e.g. `vlc --rtsp-tcp`) causes segmentation fault.

![image](https://user-images.githubusercontent.com/7534274/87253479-d2825600-c47b-11ea-86ee-adedd353a1f2.png)

Using only UDP causes packet loss and damaged frames. Even when my B800 camera is directly connected to Ethernet port of my PC. I often see such Image. Since my picture is in 4K, it causes huge traffic. Maybe other models with HD stream only are working good.

![image](https://user-images.githubusercontent.com/7534274/87253503-12493d80-c47c-11ea-9b5d-2be830e349d3.png)

I was able to mitigate flickering by adjusting some parameters in VLC:

```
vlc --network-caching=800 --no-sout-audio --rtsp-frame-buffer-size=800000 --no-skip-frames --no-drop-late-frames --udp-timeout 15
```

But it's not definitley gone. At night, when image switches to only grayscale, everything works good, so it must be bandwdth problem. I guess, TCP would solve this problem, or am I something missig, how could I solve this problem?

m1k1o:
ffmpeg log:
![image](https://user-images.githubusercontent.com/7534274/87254659-54767d00-c484-11ea-9bdf-ddfe42a44447.png)

vlc log:
![image](https://user-images.githubusercontent.com/7534274/87254668-6526f300-c484-11ea-9345-98e2f47d122f.png)

QuantumEntangledAndy:
I thought it was tcp, it certainly seems to be from my `netstat` output

```
# sudo netstat -nlp | grep neolink
tcp        0      0 127.0.0.1:8554          0.0.0.0:*               LISTEN      312/neolink
```

Are you transcoding? ffmpeg may be dropping frames because the transcoding is taking too long and the buffer maxes out. Perhaps you can try with `-c:v copy`.

m1k1o:
I'm not transcoding, using `-c:v copy`. When it is TCP, asy ou are saying, then it's weird, that `--rtsp-tcp` causes segmentation fault.

QuantumEntangledAndy:
Yes I think it's odd, I use `-rtsp_transport` in my ffmpeg command without issue and `netstat` says its using tcp not udp. Perhaps you check your own netstat to check for tcp or udp maybe your setup is somewhat different? Still perhaps we should wait for thirtythreeforty :)

QuantumEntangledAndy:
[...]

PR #141, opened 2021-03-09
Adding Debian Buster Dockerfile

m1k1o:
Switching from `alpine` to `debian buster` fixed my problem https://github.com/thirtythreeforty/neolink/issues/37. Now it seems to fail, when i connect with UDP (strange!) I added it as separate Dockerfile, just as another option. Also, switching to TCP improved image quality.

Maybe releated issue: https://github.com/thirtythreeforty/neolink/issues/46.

### DIFF
--- a/Dockerfile.debian
+++ b/Dockerfile.debian
@@ -1,0 +1,52 @@
+# Neolink Docker image build scripts
+# Copyright (c) 2020 George Hilliard
+# SPDX-License-Identifier: AGPL-3.0-only
+
+FROM docker.io/rust:buster AS build
+MAINTAINER thirtythreeforty@gmail.com
+
+ENV DEBIAN_FRONTEND=noninteractive
+RUN set -eux; \
+    apt-get update; \
+    apt-get install -y --no-install-recommends \
+    build-essential \
+    musl-dev gcc \
+    libgstrtspserver-1.0-dev \
+    libgstreamer1.0-dev \
+    libgtk2.0-dev \
+    libglib2.0-dev; \
+    apt-get clean -y; \
+    rm -rf /var/lib/apt/lists/* /var/cache/apt/*
+
+# Use static linking to work around https://github.com/rust-lang/rust/pull/58575
+ENV RUSTFLAGS='-C target-feature=-crt-static'
+
+WORKDIR /usr/local/src/neolink
+
+# Build the main program
+COPY . /usr/local/src/neolink
+RUN cargo build --release
+
+# Create the release container. Match the base OS used to build
+FROM debian:buster-slim
+
+ENV DEBIAN_FRONTEND=noninteractive
+RUN set -eux; \
+    apt-get update; \
+    apt-get install -y --no-install-recommends \
+    musl-dev \
+    libgstrtspserver-1.0-0 \
+    libgstreamer1.0-0 \
+    gstreamer1.0-plugins-good \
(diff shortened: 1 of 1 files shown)


Dockerfile.debian
```diff
@@ -0,0 +1,52 @@
...
+
+FROM docker.io/rust:buster AS build
+MAINTAINER thirtythreeforty@gmail.com
+
+ENV DEBIAN_FRONTEND=noninteractive
+RUN set -eux; \
+    apt-get update; \
+    apt-get install -y --no-install-recommends \
+    build-essential \
+    musl-dev gcc \
```
QuantumEntangledAndy, line 13: Probably don't need `musl-dev` on a buster.

Dockerfile.debian
```diff
@@ -0,0 +1,52 @@
...
+    libgstrtspserver-1.0-dev \
+    libgstreamer1.0-dev \
+    libgtk2.0-dev \
+    libglib2.0-dev; \
+    apt-get clean -y; \
+    rm -rf /var/lib/apt/lists/* /var/cache/apt/*
+
+# Use static linking to work around https://github.com/rust-lang/rust/pull/58575
+ENV RUSTFLAGS='-C target-feature=-crt-static'
```
QuantumEntangledAndy, line 22: Isn't this workaround only required on musl?

Dockerfile.debian
```diff
@@ -0,0 +1,52 @@
...
+RUN cargo build --release
+
+# Create the release container. Match the base OS used to build
+FROM debian:buster-slim
+
+ENV DEBIAN_FRONTEND=noninteractive
+RUN set -eux; \
+    apt-get update; \
+    apt-get install -y --no-install-recommends \
+    musl-dev \
```
QuantumEntangledAndy, line 37: Dev packages not need on publish image

QuantumEntangledAndy (changes requested):

QuantumEntangledAndy:
That's strange because of one of images I got you to test at the time was a buster. Maybe there has been an update upstream since then. Anyways glad your issue is fixed because this one has been bugging me for ages. I hope it's probably fixed now.

Dockerfile.debian
m1k1o: Without it, it didn't work.

Dockerfile.debian
m1k1o: Maybe because of this it didn't work without `musl-dev`. No idea.

Dockerfile.debian
QuantumEntangledAndy: Hmmm odd. I should have my buster image on one of my branches. I'll dig it up in the morning and see what else is different.

m1k1o:
Today I migrated my script to a different server, and tried the latest docker image. Still got the same problem. But after using this docker image, it complied flawlessly with latest code and works just right.

Just FYI, i had this script for last 1.5 years 24/7 in use and it worked flawlessly!  I would suggest to replace default `Dockerfile` with this version, as this is proven to work and fixes #37 and #46.

QuantumEntangledAndy:
If you want this merged might I recommend some changes.

1: remove the musl I can confirm they are not required (see [QuantumEntangledAndy/#20](https://github.com/QuantumEntangledAndy/neolink/pull/20))
2: Switch to `slim-buster` no need to use the larger buster image when the smaller one works

PR #151, opened 2021-04-19
Single login streams

QuantumEntangledAndy:
### Overview
This allows for multiple streams on a single login.

---

### Changes

- Subscriptions are now on message numbers not message IDs
- If the camera sends a response code other than 200 we raise an error
- `camera_main` and `camera_loop` now accept a slice of &[(name, outputs)]
- Arc mutexes used for `outputs` because of threaded mutability
- Added logout message
- Added stop video message
- Made start_camera loop abortable. This is used to stop both HD and SD threads when one of them errors, so that the connection can be reset

### Interesting Model Change

When requesting a camera stream we send the xml

```xml
<?xml version="1.0" encoding="utf-8"?>
<body>
<Preview version="1.1">
<channelId>0</channelId>
<handle>0</handle>
<streamType>mainStream</streamType>
</Preview>
</body>
```

Before we always set handle to `0`.

However if I request two streams with the same handle then the first stream will not send data. In fact the first stream will never send data again until the camera is restarted (instead responding with a 400 code)

So we now request `mainStream`: `handle=0` and `subStream`: `handle=1`

In the header there is already a u8 byte which is 0 for `mainstream` and 1 for `substream` so I presume that this `handle` is the modern version of that byte. Therefore, I have also reduced the size of `handle` to be a u8 which allows us to reuse the byte in the header.

### DIFF
--- a/crates/core/src/bc/de.rs
+++ b/crates/core/src/bc/de.rs
@@ -169,8 +169,7 @@ fn bc_modern_msg<'a, 'b>(
     };
 
     // Now we'll take the buffer that Nom gave a ref to and parse it.
-    let extension;
-    if ext_len > 0 {
+    let extension = if ext_len > 0 {
         // Apply the XML parse function, but throw away the reference to decrypted in the Ok and
         // Err case. This error-error-error thing is the same idiom Nom uses internally.
         let parsed = Extension::try_parse(processed_ext_buf)
@@ -182,10 +181,10 @@ fn bc_modern_msg<'a, 'b>(
         {
             context.in_bin_mode.insert(header.msg_num);
         }
-        extension = Some(parsed);
+        Some(parsed)
     } else {
-        extension = None;
-    }
+        None
+    };
 
     // Now to handle the payload block
     // This block can either be xml or binary depending on what the message expects.
--- a/crates/core/src/bc/model.rs
+++ b/crates/core/src/bc/model.rs
@@ -10,6 +10,8 @@ pub const MSG_ID_LOGIN: u32 = 1;
 pub const MSG_ID_LOGOUT: u32 = 2;
 /// Video and Audio Streams messages have this ID
 pub const MSG_ID_VIDEO: u32 = 3;
+/// Video and Audio Streams messages have this ID when a stop is requested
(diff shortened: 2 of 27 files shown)


src/bc_protocol.rs
```diff
@@ -263,12 +273,44 @@ impl BcCamera {
...
+        self.credentials = Some((username.to_string(), password.unwrap_or("").to_string()));
+
         Ok(device_info)
     }
 
     pub fn logout(&mut self) -> Result<()> {
         if self.logged_in {
-            // TODO: Send message ID 2
+            // Why does reolink bother with hashing the password during
+            // login if they just send it plain text during logout.....
```
thirtythreeforty, line 285: Lol, flawless

src/bc_protocol.rs
```diff
@@ -69,6 +71,9 @@ pub enum Error {
     #[error(display = "ADPCM Decoding Error")]
     AdpcmDecodingError(&'static str),
 
+    #[error(display = "Camera responded with Service Unavaliable")]
+    CameraServiceUnavaliable,
```
thirtythreeforty, line 75: Is there a semantic distinction between this and UnintelligibleReply?

twistedddx: From my understanding:
Unintellible is a reply that wasn't understood by Neolink.
It is likely a Neolink failure in the receive direction due to incompatible device or firmware.
But could be bad comms or hardware failure.

Unavailable is the device rejected the request. Neolink understood this rejection.
It may be a failure by Neolink to send a well formatted request, possibly due to incompatible device or firmware.
But could also just be normal behaviour by the device at certain times.

QuantumEntangledAndy: This one happens when neolink receives a valid BC packets but that the response code is not 200.

I think unintelligible is for when the response is not even deserialisable while this one is for when the camera says no.

For example when trying to start two talk streams the camera will respond with a non 200 code until the first client says it has finished talking.

QuantumEntangledAndy:
I am going to move these PRs into draft state until the subcommand one is in then I will update these to use subcommands

QuantumEntangledAndy:

PR #169, opened 2021-06-20
Add example docker-compose config

mpfl:
My current preference is to create docker-compose files for all my containers. This is, more or less, the configuration I use for my working neolink container.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,22 @@ docker run \
   thirtythreeforty/neolink
 ```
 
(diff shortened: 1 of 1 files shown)


QuantumEntangledAndy:
Could you perhaps add the yml tag after the opening triple back ticks so that it syntax highlighta

````
```yml
Code
```
````

mpfl:
Nice! I didn't even know that was a thing.

PR #180, opened 2021-07-26
Add missing variable f_bin_offset

entropin:
I don't know if the reference is wrong or if the var is missing, but i needed to add this to skip the errors


thirtythreeforty (approved):
Good catch, thanks!

PR #185, opened 2021-07-31
Refractor code into workspaces and sub commands

QuantumEntangledAndy:
This is the proposed idea in #154. Which is to takes the rust api (and later make a c-api) that contains the generic set of command and control for the camera and seperate it into it's own package. This should allow other project to be built using reolink camera.

It also changes the neolink binary such that:

- it links to this `neolink-core` crate.
- it uses subcommands
  The idea of subcommands is so that we can expand neolink into a general camera control cli

  Currently we add the subcommand `rtsp` which is the current pre-merge function
  ```bash
   neolink rtsp --config=config.toml
  ```

  Future commands might include
  - reboot
  - talk
  - mqtt

This is a draft PR to encourage early comments and discussion. I would prefer any issues about style and design choices to be brought up early


QuantumEntangledAndy:
So I think I have the basic structure in now and also some (in need for spell checking) documentaion.

QuantumEntangledAndy:
@thirtythreeforty This is now in a working state.

Currently it has the rtsp subcommand and a sub command to control the LED status light. The LED status light is quite a simple example to use as a base for other commands.

Do you want to review and merge this and I add other commands in seperate PRs? Or should I add a few more things?

QuantumEntangledAndy:
I've added a bit more to this. Namely the work in #187.

This include doing BcMedia entirely with nom and cookie_factory.

I've added tests for BcMedia deser too so we can test against the adpcm packets etc.

Also add talk and reboot commands.

Please review :)

QuantumEntangledAndy:
@thirtythreeforty Is there any chance this can get merged? I have a working UDP branch that is based off of this one that I'd like to PR

thirtythreeforty:
Ack thanks for working on this. Quite the laundry list :) let me review this one, might take me a day or two to get through it!

src/talk/cmdline.rs
```diff
@@ -0,0 +1,35 @@
...
+///
+/// `gst-launch` can be used to prepare this data
+#[derive(StructOpt, Debug)]
+pub struct Opt {
+    /// The path to the config file
+    #[structopt(short, long, parse(from_os_str))]
+    pub config: PathBuf,
+    /// The name of the camera to change the lights of. Must be a name in the config
+    pub camera: String,
+    /// The path to the audio file.
```
thirtythreeforty, line 17: Would it be possible to accept this via pipe/stdin from gstreamer directly? I think that prevents you from having to handle the microphone directly (unless that's a particularly desired feature) - just feed it from gstreamer

src/talk/errors.rs
```diff
@@ -0,0 +1,42 @@
+use err_derive::Error;
+use gstreamer::glib;
+
+/// The main error for the status-light subcommand
```
thirtythreeforty, line 4: I think this needs to be coalesced into a single Error type, as it seems prone to duplication.

However, this is at the top application level. I particularly like the `anyhow` crate for application top level errors like this. Perhaps using it would get rid of the giant enum altogether. We don't really care about the API surface, we just want the application to crash :)

QuantumEntangledAndy:
Thanks for having a look at it. I actually have more to go into it such as mqtt and udp and something called `externStream` which apparently is a mid quality stream present on some cameras

thirtythreeforty:
So I will say that there's at least 3 PRs worth of code in this review and that makes me slower at reviewing :)

when you do this kind of thing you can make a series of PRs, all except the first in draft state, of branches that build on each other.
- Reorganizes the code into workspace/lib crate - may need feedback changes but uncontroversial
- Split into subcommands - uncontroversial
- Implement reboot as a new subcommand - uncontroversial and a good motivating example for why subcommands should get merged
- Implement talk - may need more feedback changes or some design changes (hypothetical)

... my point here is that they're all uncontroversial, but the `talk` changes hold up the workspace change. And all the while the PR grows and grows :)

Looking at the commits I take it that it would be difficult to rebase out new features like `talk` at this point? Not the end of the world if so.

thirtythreeforty (commented):
I'm not finished reading thru, but here's a first few comments

QuantumEntangledAndy:
I can have a go at it. Maybe just clone branch and pop the changes back until a point then repeat. I will see

src/talk/errors.rs
QuantumEntangledAndy: Sure I think, errors in the library should be formally defined but outside in application level `anyhow` should be good

src/rtsp/gst.rs
```diff
@@ -84,9 +135,9 @@ impl GstOutputs {
         };
 
         let launch_aud = match self.audio_format {
-            Some(StreamFormat::ADPCM) => "! queue silent=true max-size-bytes=10485760 min-threshold-bytes=1024 ! rawaudioparse format=pcm pcm-format=s16le sample-rate=8000 num-channels=1 interleaved=true ! audioconvert ! rtpL16pay name=pay1", // DVI4 is converted to pcm in the appsrc
-            Some(StreamFormat::AAC) => "! queue silent=true max-size-bytes=10485760 min-threshold-bytes=1024 ! aacparse ! decodebin ! audioconvert ! rtpL16pay name=pay1 name=pay1",
-            _ => "! fakesink",
+            Some(StreamFormat::Adpcm(block_size)) => format!("caps=audio/x-adpcm,layout=dvi,block_align={},channels=1,rate=8000 ! queue silent=true max-size-bytes=10485760 min-threshold-bytes=1024 ! adpcmdec  ! audioconvert ! rtpL16pay name=pay1", block_size), // DVI4 is converted to pcm in the appsrc
```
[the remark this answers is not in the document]
QuantumEntangledAndy: Yes we did remove that. I needed an encoder for the talk aspect and I didn't want to write all that when we are using gstreamer anyways. Then I thought why have a custom decoder and not a custom encoder why not just use gstreamer in both.

Also when I switched to BcMedia format using nom it made it easier to exact out just the adpcm data from the other headers and that meant it was easier to pass it to gstreamer

src/talk/cmdline.rs
QuantumEntangledAndy: I can certainly handle std-in data but then I am not sure what the format it will be in is.

README.md
```diff
@@ -69,7 +69,7 @@ or your package manager on Linux.
 5. Launch Neolink from a shell, passing your configuration file:
 
     ```bash
-    neolink --config my_config.toml
```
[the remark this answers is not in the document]
QuantumEntangledAndy: I was thinking of that. We need to make the subcommands optional and have an optional `--config` that should only be `Some` when `subcommand` is `None`

Or... we could have one top level config

```
neolink --config=config.toml subcommand
```

But this means that each sub command has to have its config merged with the top level config

QuantumEntangledAndy:
You still want me to rebase into seperate PRs or should we keep going? I don't want to change too much while you are reading it though

thirtythreeforty:
Eh, at this point you can leave it. More just an observation that in my experience it's helpful to make PRs as small as possible

src/talk/cmdline.rs
QuantumEntangledAndy: I think for now though lets just implement it in gstreamer then add std-in as a seperate PR. I think we should have both methods

thirtythreeforty:
Better way to put it, I think you rebasing would be zero-sum: I'd save some time, but you'd spend that same amount of time.

QuantumEntangledAndy:
Sure I will endevaour to keep them minimal. I wasn't sure how to handle dependant PRs but I think the draft idea could work

README.md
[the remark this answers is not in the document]
QuantumEntangledAndy: So I am trying to get this to work but it is being stubborn. I think perhaps it will require the beta clap to do.

I can do it if I handle the case manually but that means we can't get claps nice error messages when they forget the `--config` or the subcommand.

I was originally thinking this

```rust
#[structopt(short, long, parse(from_os_str),required_unless("cmd"),conflicts_with("cmd"))]
pub config: Option<PathBuf>,
#[structopt(subcommand)]
pub cmd: Option<Command>,
```
But the `required_unless` and `conflicts_with` don't seem to work with sub commands. I will keep trying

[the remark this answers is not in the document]
QuantumEntangledAndy: I have this working, not perfectly as I'd like to do this from within clap. But it is does what we want

src/talk/errors.rs
QuantumEntangledAndy: I've been reading up on error libraries [here](https://blog.yoshuawuyts.com/error-handling-survey/) and learned about thiserror and other libraries. It made we wonder if there was a specific motivation for using err-derive over others?

QuantumEntangledAndy: I've gone and use anyhow and this removed the need for the error type at the application level

src/reboot/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+            info!("{}: Logging in", camera_config.name);
+            camera.login(&camera_config.username, camera_config.password.as_deref())?;
+
+            info!("{}: Connected and logged in", camera_config.name);
+
+            camera.reboot()?;
+        }
+    }
+
+    if !cam_found {
```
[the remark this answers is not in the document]
QuantumEntangledAndy: Done

QuantumEntangledAndy:
I think that covers the issue you raised is there anything else?

src/rtsp/gst.rs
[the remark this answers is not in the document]
QuantumEntangledAndy: Gstreamer always had an adpcm deencoder, in the original PR for merging audio I mentioned that we found a way to decode in gstreamer after writing it ourself. The discussion about which decoder to use is around [here](https://github.com/thirtythreeforty/neolink/issues/68#issuecomment-706770090)

The issue was that the gstreamer encoder needed to have the extra header information stripped and the block size passed explicitly (which was in the extra header). I asked twistedddx which way gave him better results and he said it was better with the custom one. However I prefer to not use it for the SLOC reduction and when having to use an encoder for the talk I think it just made sense to use the gstreamer one throughout.

PR #197, opened 2021-08-25
Add an mqtt subcommand

QuantumEntangledAndy:
#### Depends on #203

---

This PR adds an mqtt subcommand.

This will allow neolink to connect to an mqtt server.

- Neolink will post motion detection message to the server as
- **neolink/camname/status/motion**: `on/off`

```bash
neolink mqtt --config=config.toml
```

- Neolink will listen for the following commands

  - **neolink/camname/control/led**: `on/off`
  - **neolink/camname/control/ir**: `on/off/auto`
  - **neolink/camname/control/reboot**: `<message value ignored>`


petergp:
@QuantumEntangledAndy 
What is missing for this one?

Would love to have motion detection events in Home Assistant over MQTT.

Can I somehow help testing this?

QuantumEntangledAndy:
@petergp Motion detection from #203 should be merged first, there nothing particularly stopping it from getting merged per say, I think it is just us having not always having the time to work on it around the dayjobs etc

ytzelf:
Hi,
Coming from https://github.com/thirtythreeforty/neolink/issues/260
Could it be possible to add camera manipulation endpoint in addition to motion detection? for example
/camera1/streaming/start
/camera1/streaming/stop
/camera1/motion-streaming/start (if/when on-motion streaming is merged)
/camera1/motion-streaming/stop (if/when on-motion streaming is merged)
/camera1/battery

Thanks a lot!

QuantumEntangledAndy:
Most of those request don't make sense with the current design. We cannot stop streaming as neolink rtsp and neolink mqtt kinda act separately. Mqtt just serves as a command and control to the camera. Where as starting and stopping the stream would means starting and stopping the client.

The battery would be possible but since I don't own a battery camera I cannot reverse engineer those messages and have little drive to do so.

QuantumEntangledAndy:

PR #198, opened 2021-08-25
Disable Support for 03 Encryption

QuantumEntangledAndy:
#### Depends on #185

---

When we added the AES encryption protocol it was noted that the client sent the message

`03dc`

And the camera replied with

`02dd`

This is negociation of the encryption protocol. The `03dc` is signifing that the client supports encryption up to`03` and the camera replies with the one that should be actually used `02dd`.

Recently when working on UDP #91 it was found that the Argus Pro 2 replied with `03dd`. This is a different encryption to `02` and we do not know how to support it.

This PR makes neolink send `02dc` instead of `03dc` so as to not request this newer encryption.


QuantumEntangledAndy:
This one is ready

PR #199, opened 2021-08-25
Udp (Battery) Camera Support

QuantumEntangledAndy:
#### Depends on #185

---

This PR adds support for UDP based cameras such as battery camera.

- It uses nom/cookie factory to de/ser BcUdp packets
- It generalises the BcConnection to allow either udp or tcp though a `BcSource` enum
- It performs udp discovey based on UID
  - TODO: Add support for starting tcp based on UID discovery (next PR)

To use UDP the only change in the config is to set the address field to the uid

```toml
[[cameras]]
address = "MYUID"
```


QuantumEntangledAndy:
I separated this PR from the mqtt PR so that it can be merged independantly

QuantumEntangledAndy:
Also thought I'd try and describe UDP

UDP has two stages, Discovery and Transmit

### Discovery 
The discovery stage negociates the connection with the camera. 4 pieces of information must be shared for a connection to start.

1. Client ID (a random number we decided)
2. Camera ID (their random number)
3. SID (fixed for a camera)
4. MTU. Seems to be constant, but maybe should be found via mtu path discovery instead

The discovery stage is contained in `crates/core/src/bc_protocol/connection/udpconn/discover.rs`

### Transmit stage

README.md
```diff
@@ -175,6 +175,14 @@ services:
 The Docker image is "best effort" and intended for advanced users; questions
 about running Docker are outside the scope of Neolink.
 
+If you use a UDP only camera, such as a battery camera you will need to either
+use `--net=host` or setup a [macvlan](https://docs.docker.com/network/macvlan/)
+for the docker image that supports UDP broadcast.
+This is because UDP requires that udp broadcast messages are transmitted across
```
thirtythreeforty, line 181: Is it possible to detect this situation and yell? I suppose not, but I'm hopeful because I suspect it will save support requests.

thirtythreeforty: Also, nit, would suggest using the non-technical term "battery camera" first, as in:

> If you use a battery-powered camera (or other UDP-only camera)...

crates/core/src/bc_protocol/ping.rs
```diff
@@ -28,4 +28,40 @@ impl BcCamera {
...
+
+            let reply = sub_ping.rx.recv_timeout(RX_TIMEOUT)?;
+
+            if reply.meta.response_code != 200 {
+                return Err(Error::UnintelligibleReply {
+                    reply,
+                    why: "Camera responded with an error status code to the ping",
+                });
+            }
+            std::thread::sleep(wait);
```
thirtythreeforty, line 64: Why should `ping_every` form part of the public API for the core crate? Seems like a loop the application should do.

If this needs to be part of the public API, call ping from this method, don't copy paste.

crates/core/src/bcudp/crc.rs
```diff
@@ -0,0 +1,79 @@
...
+    0x88, 0x70, 0x6A, 0xF, 0xFF, 0xCA, 0x3B, 0x6, 0x66, 0x5C, 0xB, 0x1, 0x11, 0xFF, 0x9E, 0x65,
+    0x8F, 0x69, 0xAE, 0x62, 0xF8, 0xD3, 0xFF, 0x6B, 0x61, 0x45, 0xCF, 0x6C, 0x16, 0x78, 0xE2, 0xA,
+    0xA0, 0xEE, 0xD2, 0xD, 0xD7, 0x54, 0x83, 0x4, 0x4E, 0xC2, 0xB3, 0x3, 0x39, 0x61, 0x26, 0x67,
+    0xA7, 0xF7, 0x16, 0x60, 0xD0, 0x4D, 0x47, 0x69, 0x49, 0xDB, 0x77, 0x6E, 0x3E, 0x4A, 0x6A, 0xD1,
+    0xAE, 0xDC, 0x5A, 0xD6, 0xD9, 0x66, 0xB, 0xDF, 0x40, 0xF0, 0x3B, 0xD8, 0x37, 0x53, 0xAE, 0xBC,
+    0xA9, 0xC5, 0x9E, 0xBB, 0xDE, 0x7F, 0xCF, 0xB2, 0x47, 0xE9, 0xFF, 0xB5, 0x30, 0x1C, 0xF2, 0xBD,
+    0xBD, 0x8A, 0xC2, 0xBA, 0xCA, 0x30, 0x93, 0xB3, 0x53, 0xA6, 0xA3, 0xB4, 0x24, 0x5, 0x36, 0xD0,
+    0xBA, 0x93, 0x6, 0xD7, 0xCD, 0x29, 0x57, 0xDE, 0x54, 0xBF, 0x67, 0xD9, 0x23, 0x2E, 0x7A, 0x66,
+    0xB3, 0xB8, 0x4A, 0x61, 0xC4, 0x2, 0x1B, 0x68, 0x5D, 0x94, 0x2B, 0x6F, 0x2A, 0x37, 0xBE, 0xB,
+    0xB4, 0xA1, 0x8E, 0xC, 0xC3, 0x1B, 0xDF, 0x5, 0x5A, 0x8D, 0xEF, 0x02, 0x2D,
```
thirtythreeforty, line 65: Is there not a crate that has this table already? What variant of CRC is this?

crates/core/src/bcudp/model.rs
```diff
@@ -0,0 +1,91 @@
...
+//!
+
+use super::xml::*;
+
+/// Top level udp packet
+#[derive(Debug, PartialEq, Eq)]
+pub enum BcUdp {
+    /// Packet from the negotiate stage when connection info is exchanged
+    Discovery(UdpDiscovery),
+    /// Packet to acknoledge receipt of a data packet
```
thirtythreeforty, line 11:
```suggestion
    /// Packet to acknowledge receipt of a data packet
```

crates/core/src/bcudp/model.rs
```diff
@@ -0,0 +1,91 @@
...
+    Data(UdpData),
+}
+
+/// Magic for the UDP Discovery packet
+pub const MAGIC_HEADER_UDP_NEGO: u32 = 0x2a87cf3a;
+
+/// The Discovery packet is sent and received to init a connection
+#[derive(Debug, PartialEq, Eq)]
+pub struct UdpDiscovery {
+    // 4 Bytes Magic
```
thirtythreeforty, line 23:
```suggestion
    // The packet also contains these header fields not deserialized into this struct:
    // 4 Bytes Magic
```

crates/core/src/bcudp/model.rs
```diff
@@ -0,0 +1,91 @@
...
+    ///
+    /// This is negotiated during [`UdpDiscovery`] as cid for the client and did for the camera
+    ///
+    /// When receiving from the camera it will be cid
+    ///
+    /// When sending to the camera it should be did
+    ///
+    /// We use i32 because when we send the connection id to the reolink
+    /// register and then download the same connection id from the register_address
+    /// it comes back in an xml that is encoded as i32 (i.e. can be negative string)
```
thirtythreeforty, line 51: Thanks Java! /s

crates/core/src/bcudp/xml_crypto.rs
```diff
@@ -0,0 +1,34 @@
+const XML_KEY: [u32; 8] = [
+    0x1f2d3c4b, 0x5a6c7f8d, 0x38172e4b, 0x8271635a, 0x863f1a2b, 0xa5c6f7d8, 0x8371e1b4, 0x17f2d3a5,
+];
+
+pub(crate) fn decrypt(offset: u32, buf: &[u8]) -> Vec<u8> {
+    let key = XML_KEY
+        .iter()
+        .map(|i| (i + offset).to_le_bytes())
```
thirtythreeforty, line 8: Can you express the key array as bytes, so you don't need to do little-endian type conversions? You can still group them visually in the source code

dissector/udp.md
```diff
@@ -0,0 +1,315 @@
...
...BE, 0xD5, 0xE5, 0xB7, 0xEF, 0xDC, 0x7C, 0x21, 0xDF, 0xDB, 0xB, 0xD4, 0xD2, 0xD3, 0x86, 0x42, 0xE2, 0xD4, 0xF1, 0xF8, 0xB3, 0xDD, 0x68, 0x6E, 0x83, 0xDA, 0x1F, 0xCD, 0x16, 0xBE, 0x81, 0x5B, 0x26, 0xB9, 0xF6, 0xE1, 0x77, 0xB0, 0x6F, 0x77, 0x47, 0xB7, 0x18, 0xE6, 0x5A, 0x8, 0x88, 0x70, 0x6A, 0xF, 0xFF, 0xCA, 0x3B, 0x6, 0x66, 0x5C, 0xB, 0x1, 0x11, 0xFF, 0x9E, 0x65, 0x8F, 0x69, 0xAE, 0x62, 0xF8, 0xD3, 0xFF, 0x6B, 0x61, 0x45, 0xCF, 0x6C, 0x16, 0x78, 0xE2, 0xA, 0xA0, 0xEE, 0xD2, 0xD, 0xD7, 0x54, 0x83, 0x4, 0x4E, 0xC2, 0xB3, 0x3, 0x39, 0x61, 0x26, 0x67, 0xA7, 0xF7, 0x16, 0x60, 0xD0, 0x4D, 0x47, 0x69, 0x49, 0xDB, 0x77, 0x6E, 0x3E, 0x4A, 0x6A, 0xD1, 0xAE, 0xDC, 0x5A, 0xD6, 0xD9, 0x66, 0xB, 0xDF, 0x40, 0xF0, 0x3B, 0xD8, 0x37, 0x53, 0xAE, 0xBC, 0xA9, 0xC5, 0x9E, 0xBB, 0xDE, 0x7F, 0xCF, 0xB2, 0x47, 0xE9, 0xFF, 0xB5, 0x30, 0x1C, 0xF2, 0xBD, 0xBD, 0x8A, 0xC2, 0xBA, 0xCA, 0x30, 0x93, 0xB3, 0x53, 0xA6, 0xA3, 0xB4, 0x24, 0x5, 0x36, 0xD0, 0xBA, 0x93, 0x6, 0xD7, 0xCD, 0x29, 0x57, 0xDE, 0x54, 0xBF, 0x67, 0xD9, 0x23, 0x2E, 0x7A, 0x66, 0xB3, 0xB8, 0x4A, 0x61, 0xC4, 0x2, 0x1B, 0x68, 0x5D, 0x94, 0x2B, 0x6F, 0x2A, 0x37, 0xBE, 0xB, 0xB4, 0xA1, 0x8E, 0xC, 0xC3, 0x1B, 0xDF, 0x5, 0x5A, 0x8D, 0xEF, 0x02, 0x2D,];
```
thirtythreeforty, line 46: Please don't paste this here, just say "refer to [file X]". If possible, reference an open source crate - surely this CRC algorithm isn't unique? (Although, it _is_ Reolink we're talking about here...)

thirtythreeforty: (also below for the encryption)

dissector/udp.md
```diff
@@ -0,0 +1,315 @@
+# BcUDP
+
+This document describes the UDP protocol. UDP, unlike TCP, is lossy, there
```
thirtythreeforty, line 3: AWESOME docs, nice work

sample_config.toml
```diff
@@ -50,7 +50,13 @@ name = "storage shed"
 username = "admin"
 password = "987654321"
 address = "192.168.1.245:9000"
+# The address could instead be set to a camera UID like this
+# address = "ABCD01234567890EFG"
```
thirtythreeforty, line 54: I would prefer to introduce a new key, `uid`. They are technically separate namespaces - it's not _totally_ [inconceivable](https://www.youtube.com/watch?v=D9MS2y2YU_o) that someone could give `ABCD01234567890EFG` a name in DNS.

Then, we can start recommending people use the uid of their cameras, and only use `address` if they're on a different subnet or like Docker for some reason :)

sample_config.toml
```diff
@@ -50,7 +50,13 @@ name = "storage shed"
...
 
+# If you use an NVR that relays several camera connections you can choose which
+# camera to connect to with by setting the `channel_id`
+#
 # By default channel_id = 0. Eg the first connected camera on the device
-# Note this counts from 0. An 8 channel NVR would have channels 0 through 7
+# **Note**: that unlike in the offical client the  numbering starts from 0 not 1.
+# An 8 channel NVR would have channels 0 through 7
 # channel_id = 1
```
thirtythreeforty, line 62:
```suggestion
# channel_id = 0
```

thirtythreeforty: (defaults as comments in sample config files are a common pattern)

thirtythreeforty (commented):
Haven't finished reading through the UDP implementation itself, so more comments to come - quite a lot of SLoC - but this is really cool. Honestly I think this will make Neolink a much bigger deal than it is now, because a lot more people have battery-powered cams than do the PoE models, since it's a lower barrier to entry.

sample_config.toml
thirtythreeforty: (because you have to admit, the real client's workflow for discovering and adding cameras is pretty slick)

crates/core/src/bc_protocol/ping.rs
QuantumEntangledAndy: I think perhaps remove this, it was part of my testing because the UDP camera needs a new message from the client regularly or it will disconnect. This is now done through the keep alive message in `BcConnection`

crates/core/src/bcudp/crc.rs
QuantumEntangledAndy: I have no idea what the CRC is, most of this code comes from the python program [here](https://github.com/jdillenkofer/camera_proxy) I had a look into this but it didn't seem to match the CRC uses for UDP protocol and CRCs are not my area of expertise so I am usure where it is from.

crates/core/src/bcudp/xml_crypto.rs
QuantumEntangledAndy: Again this comes from [here](https://github.com/jdillenkofer/camera_proxy) I tried to get it to do it in just bytes but it kept failing to work right, so I just got it working in the same way it was implemented there thinking to come back when I have time.

crates/core/src/bcudp/crc.rs
QuantumEntangledAndy: Did some google and maybe [this](https://github.com/srijs/rust-crc32fast/blob/master/src/table.rs) one will work :)

QuantumEntangledAndy: Tried every algorithm in the CRC crate.

The current method provides this:
- Should BE: 1115833684

While every u32 method in the crc crate provides a different value:
- CRC_32_ISCSI was 2699336615
- CRC_32_AIXM was 103931297
- CRC_32_AUTOSAR was 2232806661
- CRC_32_BASE91_D was 1196129179
- CRC_32_BZIP2 was 2046554463
- CRC_32_CD_ROM_EDC was 419968025
- CRC_32_CKSUM was 3152376032
- CRC_32_ISO_HDLC was 1090306792
- CRC_32_JAMCRC was 3204660503
- CRC_32_MPEG_2 was 2248412832
- CRC_32_XFER was 3116207526

QuantumEntangledAndy: Ok I i figured it out. It's a crc with a polynomial of 0x04c11db7 like CRC_32_ISO_HDLC BUT it has an init value of 0x00000000 (instead of 0xffffffff) and an xorout of 0x00000000 (instead of 0xffffffff)

Two ways to implement this:

1. Use the crc crate with the poly and parameters manually supplied.
2. Use the fastcrc crate but adjust the init value and undo the final xorout

QuantumEntangledAndy: I will use the fastcrc crate as it:
- uses a fixed table lookup (instead of calculating from the polynomial)
- Uses sse or aarch64 acceletation were possible and falls back to 4 bytes at a time were not

dissector/udp.md
QuantumEntangledAndy: I have tried to describe the crc polynomial and conditions which should make it possible to find the right algorithm for any language. I could also link to the [crc catalogue](https://reveng.sourceforge.io/crc-catalogue/all.htm) which discusses what these parameters mean, but there is no exact match for this crc algorithm

dissector/udp.md
QuantumEntangledAndy: Thanks, it was helpful during develop, to have this at hand, as it was too complex for me to remember all the details as I went

crates/core/src/bc_protocol/ping.rs
QuantumEntangledAndy: Removed

README.md
QuantumEntangledAndy: Done

crates/core/src/bcudp/xml_crypto.rs
QuantumEntangledAndy: Perhaps my bitwise operators are not good or something, but how do I do a byte wise addition of two u32s if I break it up into u8s? The `offset` is added to the `XML_KEY` which can make it carry the value to the next byte. I could represent it as u8 bytes then, loop them four at a time while reconstucting the u32 for the addition, but this is visually more complex then just the `to_le_bytes`

QuantumEntangledAndy: I could also assume no carry such that the last bit of the byte is zero. But I might need to adjust how the offset is made in other parts of the code

sample_config.toml
QuantumEntangledAndy: What should we do in the event that both uid and addr are given? Error, use addr, use uid?

QuantumEntangledAndy: Making it do both in a decent way might require a larger refractor. Ideally I'd want it to connect to both, then drop one like the offical client does. But it's really not set up to do that.

thirtythreeforty: Hmm, a good question. The strictest thing to do right now might be to error in that situation - that lets us open up the syntax in the future to allow what you describe, but lets us avoid breaking backward compat with config files written today.

crates/core/src/bcudp/xml_crypto.rs
thirtythreeforty: Oh I see. You're correct, it needs to be u32 because of the addition!

crates/core/src/bcudp/crc.rs
```diff
@@ -1,79 +1,7 @@
...
-        let val: u32 = TABLE[offset + 3] << 24
-            | TABLE[offset + 2] << 16
-            | TABLE[offset + 1] << 8
-            | TABLE[offset];
-        r = val ^ (r >> 8);
-    }
-    r
+    let mut hasher = Hasher::new_with_initial(0xffffffff);
+    hasher.update(payload);
+    hasher.finalize() ^ 0xffffffff_u32
```
thirtythreeforty, line 6: Thank you for digging into this!

QuantumEntangledAndy: I'm just glad bc didn't entirely role their own. It's a crc not in the crc-catalogue, the closest is `CRC-32/CKSUM` but with the difference that bc skipped the final xor.

sample_config.toml
QuantumEntangledAndy: Done we error out for now. This is done in the validation. I handle the camera address logic in a `utils.rs` which will helpfully ease the transition when we get to it as it is all in one place for all subcommands

crates/core/src/bc_protocol/connection/udpconn/discover.rs
```diff
@@ -0,0 +1,543 @@
...
+
+        Ok(UdpDiscover {
+            address: camera_address,
+            client_id,
+            camera_id,
+            mtu,
+        })
+    }
+
+    fn get_register(
```
QuantumEntangledAndy, line 174:
```suggestion
    // This function will contact the p2p relay servers
    //
    // It will ask each of the servers for details on a specific UID
    //
    // On success it returns the M2cQr that the p2p relay
    // server has about the UID
    fn get_register(
```

crates/core/src/bc_protocol/connection/udpconn/discover.rs
```diff
@@ -0,0 +1,543 @@
...
+                        match BcUdp::deserialize(&buf[0..bytes_read]) {
+                            Ok(BcUdp::Discovery(UdpDiscovery {
+                                tid: _,
+                                payload:
+                                    UdpXml {
+                                        m2c_q_r: Some(m2c_q_r),
+                                        ..
+                                    },
+                            })) if m2c_q_r.reg.port == 0 || m2c_q_r.reg.ip.is_empty() => {
+                                // This register is empty
```
QuantumEntangledAndy, line 220:
```suggestion
                                // This register is empty
                                // Sometimes a registery will return a M2cQr but it's port and addr
                                // will be 0
                                // In this case we ignore this server and contact another
```

crates/core/src/bc_protocol/connection/udpconn/discover.rs
```diff
@@ -0,0 +1,543 @@
...
+                    // This p2p server didn't have the data we need try another
+                    break;
+                }
+            }
+        }
+        debug!("All registers tried");
+        Err(Error::ConnectionUnavaliable)
+    }
+
+    fn discover_from_uuid_remote(
```
QuantumEntangledAndy, line 253:
```suggestion
    // This will start remote discovery against the reolink p2p servers
    //
    // It has the following stages
    // - Contact the p2p register: This will get the register, relay and log server ip addresses
    // - Register the client ip address to the register (This can probably be used for NAT holepunching if the camera
    //    tries to connect to us using this data, currently we don't register a real address though, just `0.0.0.0`)
    // - The register will return the camera ip/port and SID
    // - Connect to the camera using register's provided ip/port and negotiate a cid and did
    // - Send a message to the reolink log server that we will connect locally
    // - Send a message to the camera that we could connect remotely
    fn discover_from_uuid_remote(
```

crates/core/src/bc_protocol/connection/udpconn/discover.rs
```diff
@@ -0,0 +1,543 @@
...
+        debug!("Registering this address: {:?}", local_addr);
+
+        let local_family = if local_addr.ip().is_ipv4() { 4 } else { 6 };
+        let msg = BcUdp::Discovery(UdpDiscovery {
+            tid,
+            payload: UdpXml {
+                c2r_c: Some(C2rC {
+                    uid: uid.to_string(),
+                    cli: IpPort {
+                        ip: "10.251.254.10".to_string(),
```
QuantumEntangledAndy, line 286: Should not be fixed but should be client ip address

crates/core/src/bc_protocol/connection/udpconn/transmit.rs
```diff
@@ -0,0 +1,340 @@
...
+        results
+    }
+
+    // Revieve a new packet only update if entry is new
+    fn receive(&self, packet_id: u32, payload: Vec<u8>) {
+        let mut locked = self.buffer.lock().unwrap();
+        let _ = locked.entry(packet_id).or_insert(payload);
+    }
+
+    fn get_needed_packets(&self) -> Option<(u32, Vec<u8>)> {
```
QuantumEntangledAndy, line 177:
```suggestion
    // This gets the needed payload to form a `UdpAck` packet
    fn get_needed_packets(&self) -> Option<(u32, Vec<u8>)> {
```

crates/core/src/bc_protocol/connection/udpconn/transmit.rs
```diff
@@ -0,0 +1,340 @@
...
+
+        // Find the last contigous packet
+        while locked.contains_key(&(start + 1)) {
+            start += 1;
+        }
+
+        // Find last packet in buffer
+        let vec = if let Some(end) = locked.keys().max() {
+            let mut vec = vec![];
+            for i in (start + 1)..(end + 1) {
```
QuantumEntangledAndy, line 193:
```suggestion
            // From last contiguous packet to last recieved packet
            // create a payload of `00` (unreceived) and `01` (received)
            // that can be used to form the `UdpAck` packet
            for i in (start + 1)..(end + 1) {
```

crates/core/src/bc_protocol/connection/udpconn/transmit.rs
```diff
@@ -0,0 +1,340 @@
...
+                        payload:
+                            UdpXml {
+                                d2c_disc: Some(D2cDisc { cid, did }),
+                                ..
+                            },
+                        ..
+                    }) if cid == discovery_result.client_id
+                        && did == discovery_result.camera_id =>
+                    {
+                        // Reply with C2D_Disc and end
```
QuantumEntangledAndy, line 239:
```suggestion
                        // Camera sent a disconnect
                        // Reply with C2D_Disc and end
```

crates/core/src/bc_protocol/connection/udpconn/transmit.rs
```diff
@@ -0,0 +1,340 @@
...
+                    BcUdp::Ack(UdpAck { .. }) => {
+                        debug!("Got a UdpAck for another client id");
+                    }
+                    BcUdp::Data(UdpData {
+                        connection_id: cid,
+                        packet_id,
+                        payload,
+                    }) if cid == discovery_result.client_id => {
+                        // Seperated as let/if due to clippy recommend
+                        trace!("Recieving UdpData packet {}", packet_id);
```
QuantumEntangledAndy, line 263:
```suggestion
                    }) if cid == discovery_result.client_id => {
                        trace!("Recieving UdpData packet {}", packet_id);
```

crates/core/src/bcudp/de.rs
```diff
@@ -0,0 +1,369 @@
...
+    let (buf, _unknown_a) = verify(le_u32, |&x| x == 0)(buf)?;
+    let (buf, _unknown_b) = verify(le_u32, |&x| x == 0)(buf)?;
+    let (buf, packet_id) = le_u32(buf)?; // This is the point at which the camera has contigious
+                                         // packets to
+    let (buf, _unknown) = le_u32(buf)?;
+    let (buf, payload_size) = le_u32(buf)?;
+    let (buf, payload) = if payload_size > 0 {
+        let (buf, t_payload) = take!(buf, payload_size)?; // It is a binary payload of
+                                                          // `00 01 01 01 01 00 01`
+                                                          // Probably a truth map of missing packets
```
QuantumEntangledAndy, line 145:
```suggestion
                                                          // This is a truth map of missing packets
```

dissector/udp.md
```diff
@@ -0,0 +1,285 @@
...
+                           It is also used for the encryption.
+- 4 Bytes Checksum: The checksum of the payload encrypted payload
+
+The checksum can be calculated with a polynomial of `0x04c11db7`, an init value
+of `0x00000000` and an  xorout of `0x00000000`.
+
+
+## Payload
+
+- The XML payloads and encrypted with a simple xor algorithm. See the full
```
QuantumEntangledAndy, line 49:
```suggestion
- The XML payloads are encrypted with a simple xor algorithm. See the full
```

dissector/udp.md
```diff
@@ -0,0 +1,285 @@
...
+- **cid**: The connection ID of the client
+- **did**: The connection ID of the camera
+- **mtu**: The maximum transmission unit of the connection. Which is the
+           largest packet size in bytes
+
+
+## T Payload Camera
+
+Camera replies with this payload on the port specified in `C2D_C`
+
```
QuantumEntangledAndy, line 174:
````suggestion
Camera replies with this payload on the port specified in `C2D_C`

```xml
<P2P>
<D2C_T>
<sid>62097899</sid>
<conn>local</conn>
<cid>82000</cid>
<did>528</did>
</D2C_T>
</P2P>
```
````

File: dissector/udp.md [hunk 24]
```diff
@@ -0,0 +1,285 @@
...
+- **sid**: ID of the camera
+- **conn**: Type of connection only observed `local` value
+- **rsp**: Unknown always 0
+- **cid**: The connection ID of the client
+- **did**: The connection ID of the camera
+- **time_r**: Should be the time but it is always 0
+
+# UDP Ack
+
+These messages are sent to acknowledge receipt of message. They are header only.
```
QuantumEntangledAndy, line 224:
```suggestion
These messages are sent to acknowledge receipt of message
```

dissector/udp.md
```diff
@@ -0,0 +1,285 @@
...
+- 4 Bytes unknown: Always `00000000` for for UDP Ack
+- 4 Bytes unknown: Always `00000000` for for UDP Ack
+- 4 Bytes Last Packet ID: Last received packet id
+- 4 Bytes Unknown: Observed values `00000000`, `d6010000`, `d7160000` `09e00000` **NEEDS INFO**
+- Payload Size
+- Paload
+
+**To Investigate:**
+  - Why does the unknown byte change. It starts at zero and remains that way
+    for a second. Then seems to change and remain at the new value for another second
```
QuantumEntangledAndy, line 239:
```suggestion
    for a second. Then seems to change and remain at the new value for another second.
    This may be some metric such as average ttl for the last second, or some calculation
    of the jitter
```

Subtixx:
Using an Reolink Argus 3 results in a parsing error:
```
[2021-09-11T14:28:33Z ERROR neolink_core::bc_protocol::connection::bcconn] Deserialization error: Deserialization error
[2021-09-11T14:28:33Z ERROR neolink_core::bc_protocol::connection::bcconn] caused by: Parsing error
```

QuantumEntangledAndy:
@Subtixx this is a known error that is dealt with in #198 once that gets merged it should be fine.

Argus 3 uses a newer encryption algorithm that we need to account for.

Subtixx:
Thanks. So is there any neolink binary that I can use for the Argus 3 right now?

QuantumEntangledAndy:
Yeah my `udp_docker` branch should work.

Subtixx:
It doesn't but oh well (says auth failed even though the credentials are correct, even verified by logging out of the app & back in). Can't report it anywhere, and this is the wrong place to do this.

QuantumEntangledAndy:
Maybe open an issue on my fork and we can discuss it there. It's getting late in my timezone now so won't be able to help much until tomorrow though.

MasterPlexus:
Hi. Is this also part of the current downloadable binary for armhf? As far as I understand this should also work with Argus 2, but at the moment I'm not able to use it for that Cams..

QuantumEntangledAndy:
@MasterPlexus Can you open a new issue and give up some more details such as your log files when in debug mode.

```bash
export RUST_LOG=debug
```

PR #206, opened 2021-09-01
Fix IFrame and PFrame identification for Swann DVRs

troyrollo:
The Swann DVRs (at least) include the channel number of the streamed channel in (at least) IFrame and PFrame magic numbers. Without this patch, there is a Parsing error leading to a Deserialisation error if a channel ID other than zero is used.

With this patch, neolink works with my Swann DVR-4550 and presumably many other Swann DVRs.

### DIFF
--- a/crates/core/src/bcmedia/de.rs
+++ b/crates/core/src/bcmedia/de.rs
@@ -76,8 +76,8 @@ fn bcmedia(buf: &[u8]) -> IResult<&[u8], BcMedia> {
             *x,
             MAGIC_HEADER_BCMEDIA_INFO_V1
                 | MAGIC_HEADER_BCMEDIA_INFO_V2
-                | MAGIC_HEADER_BCMEDIA_IFRAME
-                | MAGIC_HEADER_BCMEDIA_PFRAME
+                | MAGIC_HEADER_BCMEDIA_IFRAME..=MAGIC_HEADER_BCMEDIA_IFRAME_LAST
+                | MAGIC_HEADER_BCMEDIA_PFRAME..=MAGIC_HEADER_BCMEDIA_PFRAME_LAST
                 | MAGIC_HEADER_BCMEDIA_AAC
                 | MAGIC_HEADER_BCMEDIA_ADPCM
         )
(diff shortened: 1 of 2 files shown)


QuantumEntangledAndy:
Interesting. @twistedddx is our swann camera expert, I'm surprised he hasn't seen this one yet. Do you know what models this would affect?

twistedddx:
DVR instead of NVR means analog not IP cameras. The DVR does all IP and h264 encoding etc.
I assume the DVR behaves like a NVR towards the client.

This means the first 0 in 00dc and 01dc is channel number?
This patch changes it to 91dc, is that channel_id 9 meaning camera 10?
I assume this is a 16 channel DVR, is channel_id 10 A1DC?
Does it do the same for the audio data eg 91wb?

Maybe only DVR's show the channel number as described here.
My NVR does not.

The DVR8-4550 and similar channel count firmware looks nothing like the packages for equipment I have seen from Swann or Reolink before.
I didn't realize Swann even had any BC like products still for sale.
Previously it appeared when Reolink began Swann abandoned all their BC products with no further app or firmware updates.

twistedddx:
Perhaps the 0 originally did mean channel_id 0 but with IP cameras the BCMedia stream is created by the camera itself which only has channel_id 0.
The NVR may just relay that stream without changing it causing all streams to have the channel_id 0 in that field.
The DVR probably is more correctly following the original specification.

troyrollo:
At a guess, the potentially affected devices would be the ones compatible with the SwannView Link app, according to Swann: 460, 470, 480, 485, EasyView, GuardianEye, 1260, 1425, 1450, 1580, 3100, 3200, 3250, 3425, 3450, 4100, 4200, 4400, 4550, 4600, 4750, 7072, 7075, 7082, 7085, 7285, 7300, 7400, DVR-8000, 8100, 4ch 8200 Series DVRs & NVRs

QuantumEntangledAndy:
Do you know if these camera have sound I think perhaps we should change the adpcm and aac to a range too. Also I think we should do the whole channel range from 1-16, not everyone would raise an issue if it couldn't handle channels > 10, and I'd rather not inconvenience them.

QuantumEntangledAndy:
I also wonder if there's any channels encoded in the 1001 info packets.

twistedddx:
I have a nvr-7300 / nvr-7400 and it does not do this.
I think it may be unique to DVR.

troyrollo:
The DVR-4550 only records audio through a separate audio input and only makes it available through channel 1, so I am not in a position to say what will happen with audio on another channel.

For channels above 10 we would only be guessing at what would happen. Perhaps they treat the channel number as a hexadecimal digit (but the codes are not hexadecimal digits as demonstrated by the "wb" for audio), or they could just take the lazy approach given that it is a binary protocol and continue into the symbols following the ASCII digits. Rather than guess, the answer may be to mask off the channel byte altogether on the (likely reasonable) assumption that no other type of data will be found with the "dc" portion of the code.

Taking that approach however means it will probably just work, with technically incorrect logic, so that there will be no reason at all for somebody with a 16 channel DVR to report anything and the technically incorrect logic will remain forever.

The third option would be to generate an error message displaying the code actually seen, requesting that this be reported so that the software can be updated to support it.

QuantumEntangledAndy:
It's ok to leave it as is I was mostly just curious about the audio, but it seems to be handled differently so we can't really observe it.

Anyways thanks for the PR I hope others will find it helpful :). Sorry for being a bit absent the last few days it's been busy at work, so haven't been able to discus this properly :(

QuantumEntangledAndy (approved):

QuantumEntangledAndy:
If you have any wireshark dumps you think might be appropiate to put into the intergration test please send them and I will add it in

PR #203, opened 2021-08-31
Motion Detection

QuantumEntangledAndy:
This PR adds motion detection to `neolink_core`

- It also contains some small fixes to documentation and comments

- Additionally, while writing the motion detection code I made the callback be able to gracefully shutdown the motion detection loop returning `Ok(false)`.
  - I thought this was a nice feature so I added it to the stream callback too (although it is mostly unused in our own code it may be of use as part of the library)
  - Let me know if you want me to drop this

PR #210, opened 2021-09-15
Motion based Stream

QuantumEntangledAndy:
Depends on #199 #203

---

This PR intents to add motion based streaming.

### Features

- [x] When motion is stopped it can show the last picture
- [x] When motion is stopped it can show a black screen
- [x] Configurable timeout

The intent behind this is to reduce battery consumption for battery cameras, however this PR may not be enough as we still need to be logged in to check the motion status. If this is the case we may need implement an smtp server and set the camera to send us message through that when motion starts.


QuantumEntangledAndy:
@thirtythreeforty I am at a bit of an im-pass on this code. I can make it do motion based streams using [input-selector](https://gstreamer.freedesktop.org/documentation/coreelements/input-selector.html) and [imagefreeze](https://gstreamer.freedesktop.org/documentation/imagefreeze/index.html) however it requires that I decode the video to raw format putting strain on the cpu.

It also causes artifacts when swapping the input selector from the camera stream to the paused stream. I think could fix this by decoding always to raw format and reencoding rather than only encoding the paused stream.

As an alternative I could just stop passing input to gstraemer when the camera is in the paused state. Already running client streams will either pause until buffers are avaliable or stop (depending on the client). However during this time no new client can connect.

Do you have any recommendation as to how this should be designed?

- Do full re-encode for more features
- Just stop sending data when camera is in paused state

QuantumEntangledAndy:
I believe this is working now. Here is the config I am using for it:

```toml
[[cameras]]
name = "Cammy"
username = "****"
password = "****"
address = "****:9000"
  [cameras.pause]
  on_motion = false
  on_client = false
  mode = "none"
  timeout = 1.0
```

- When `on_motion` is true the camera will pause streaming when motion is stopped and resume it when motion is started
- When `on_client` is true the camera will pause while there is no client connected.
  - This was deceptivly tricky as there is not an easy way to tell if these is a client connected in gstreamer. It was finally done by tracking the state of the app_src. If the app_src is destroyed then there is no client
- `timeout` handels how long to wait after motion stops before pausing the stream
- `mode` has the following values:
  - "black": Switches to a black screen with full reencode
  - "still": Switches to a still image with full reencode
  - "test": Switches to the gstreamer tet image with full reencode
  - "none": This one is interesting and is the default. This does not reencode at all. It just resends the last iframe the camera sents into the app_src. **I think for most use cases this one should be used as it has the least effort on the cpu and gives what you would expect**

QuantumEntangledAndy:

PR #249, opened 2022-01-20
Module video4linux

0ry0n:
I thought about introducing a module for v4l support in order to send the video payload to a /dev/video* device created via [v4l2loopback](https://github.com/umlaeute/v4l2loopback).

This way you would get the video stream without using the "heavy" gstreamer, useful when you need the video stream locally. Example: (raspberry: neolink -> NVR software).

Unfortunately I don't know rust :(. I tried to implement it trying to follow the structure of the rtsp module, but there is some problem.

### DIFF
--- a/crates/core/src/bc/model.rs
+++ b/crates/core/src/bc/model.rs
@@ -16,6 +16,10 @@ pub const MSG_ID_TALKABILITY: u32 = 10;
 pub const MSG_ID_TALKRESET: u32 = 11;
 /// Reboot messages have this ID
 pub const MSG_ID_REBOOT: u32 = 23;
+/// Request motion detection messages
+pub const MSG_ID_MOTION_REQUEST: u32 = 31;
+/// Motion detection messages
+pub const MSG_ID_MOTION: u32 = 33;
 /// Version messages have this ID
 pub const MSG_ID_VERSION: u32 = 80;
 /// Getting PIR status messages have this ID
--- a/crates/core/src/bc/xml.rs
+++ b/crates/core/src/bc/xml.rs
@@ -69,6 +69,9 @@ pub struct BcXml {
     /// Revieced as part of the TalkAbility request
     #[yaserde(rename = "TalkAbility")]
     pub talk_ability: Option<TalkAbility>,
+    /// Received when motion is detected
+    #[yaserde(rename = "AlarmEventList")]
+    pub alarm_event_list: Option<AlarmEventList>,
 }
 
 impl BcXml {
@@ -410,6 +413,35 @@ pub struct AudioConfigList {
     /// The supported audio configs
     #[yaserde(rename = "audioConfig")]
     pub audio_config: AudioConfig,
+}
+
+/// An XML that desctibes a list of events such as motion detection
+#[derive(PartialEq, Eq, Default, Debug, YaDeserialize, YaSerialize)]
+pub struct AlarmEventList {
+    /// XML Version
+    #[yaserde(attribute)]
+    pub version: String,
+    /// List of events
+    #[yaserde(rename = "AlarmEvent")]
+    pub alarm_events: Vec<AlarmEvent>,
+}
+
+/// An alarm event. Camera can send multiple per message as an array in AlarmEventList.
+#[derive(PartialEq, Eq, Default, Debug, YaDeserialize, YaSerialize)]
+pub struct AlarmEvent {
+    /// XML Version
+    #[yaserde(attribute)]
+    pub version: String,
+    /// The channel the event occured on. Usually zero unless from an NVR
+    #[yaserde(rename = "channelId")]
+    pub channel_id: u8,
+    /// Motion status. Known values are `"MD"` or `"none"`
+    pub status: String,
+    /// The recording status. Known values `0` or `1`
+    pub recording: i32,
+    /// The timestamp associated with the recording. `0` if not recording
+    #[yaserde(rename = "timeStamp")]
+    pub timeStamp: i32,
 }
 
 /// Convience function to return the xml version used throughout the library
--- a/crates/core/src/bc_protocol.rs
+++ b/crates/core/src/bc_protocol.rs
@@ -11,6 +11,7 @@ mod errors;
 mod ledstate;
 mod login;
 mod logout;
+mod motion;
 mod ping;
 mod pirstate;
 mod reboot;
@@ -27,6 +28,7 @@ pub use errors::Error;
 pub use ledstate::LightState;
 pub use pirstate::PirState;
 pub use resolution::*;
+pub use motion::{MotionOutput, MotionOutputError, MotionStatus};
 pub use stream::{Stream, StreamOutput, StreamOutputError};
 
 type Result<T> = std::result::Result<T, Error>;
--- a/crates/core/src/bc_protocol/connection/filesub.rs
+++ b/crates/core/src/bc_protocol/connection/filesub.rs
@@ -24,10 +24,7 @@ pub struct FileSubscriber {
 }
 
 impl FileSubscriber {
-    /// Creates a binary subsciber from a BcSubscrption.
-    /// When reading the next packet it will skip over multiple
-    /// Bc packets to fill the binary buffer so ensure you
-    /// only want binary packets when calling read
+    /// Creates a file subsciber from a list of files
     pub fn from_files<P: AsRef<Path>>(paths: Vec<P>) -> FileSubscriber {
         FileSubscriber {
             files: paths
--- a/crates/core/src/bc_protocol/ledstate.rs
+++ b/crates/core/src/bc_protocol/ledstate.rs
@@ -3,7 +3,7 @@ use crate::bc::{model::*, xml::*};
 
 impl BcCamera {
     /// Get the [LedState] xml which contains the LED status of the camera
-    pub fn get_ledstate(&mut self) -> Result<LedState> {
+    pub fn get_ledstate(&self) -> Result<LedState> {
         let connection = self
             .connection
             .as_ref()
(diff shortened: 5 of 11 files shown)


src/config.rs
```diff
@@ -73,6 +75,16 @@ pub(crate) struct CameraConfig {
     #[validate(range(min = 0, max = 31, message = "Invalid channel", code = "channel_id"))]
     #[serde(default = "default_channel_id")]
     pub(crate) channel_id: u8,
+
+    #[validate(range(min = 0, max = 31, message = "Invalid device number", code = "v4ldevice"))]
+    pub(crate) v4ldevice: u8,
```
QuantumEntangledAndy, line 80: This means we must all supply a number for all camera I would recommend it be a `Option<u8>` with a default value of `None` so that it won't create a v4l if its not in the config.

src/v4l/mod.rs
```diff
@@ -0,0 +1,249 @@
...
+            // syntax for that.
+            let arc_cam = Arc::new(camera);
+
+            if ["mainStream"].iter().any(|&e| e == arc_cam.v4lstream) {
+                let v4l = V4lDevice::new(arc_cam.v4ldevice as usize);
+                let mut outputs = v4l
+                    .add_stream()
+                    .unwrap();
+                let main_camera = arc_cam.clone();
+                s.spawn(move |_| camera_loop(&*main_camera, Stream::Main, &mut outputs, true));
```
QuantumEntangledAndy, line 67: When you make `arc_cam.v4ldevice` an `Option<u8>` as suggested above you'd need to make this conditional like this:

```rust
if let Some(v4ldevice) = arc_cam.v4ldevice {
  let v4l = V4lDevice::new(v4ldevice as usize);
  let mut outputs = v4l
      .add_stream()
      .unwrap();
  let main_camera = arc_cam.clone();
  s.spawn(move |_| camera_loop(&*main_camera, Stream::Main, &mut outputs, true));
}
```

QuantumEntangledAndy: For the other ones too, of just one big if outside all of them

src/v4l/v4lt.rs
```diff
@@ -0,0 +1,148 @@
...
+                // since we control the input buffer - so just enforce that size on the output buffer.
+                let buf_out = &mut buf_out[0..payload.data.len()];
+
+                buf_out.copy_from_slice(&payload.data);
+                buf_out_meta.field = 0;
+                //buf_out_meta.bytesused = buf_in_meta.bytesused;
+            }
+            _ => {
+                //Ignore other BcMedia
+            }
```
QuantumEntangledAndy, line 80: You may want to also listen to the `BcMediaInfoV2` and `BcMediaInfoV1` packets. They contain information such as the width and height. Check out `crates/core/src/bcmedia/model.rs`. These messages are only sent once at the beginning of a stream so you'll have to cache the results in your `V4ltOutputs`

src/v4l/v4lt.rs
```diff
@@ -0,0 +1,148 @@
...
+    fn apply_format(&self) {
+        let launch_vid = match self.video_format {
+            Some(StreamFormat::H264) => {
+                b"AVC1"
+            }
+            Some(StreamFormat::H265) => {
+                b"HEVC"
+            }
+            // TODO
+            _ => b"AVC1",
```
QuantumEntangledAndy, line 119: Perhaps you should just change this to `unreachable!()` since you have already filtered this out with the `set_format`

src/v4l/v4lt.rs
```diff
@@ -0,0 +1,148 @@
...
+        };
+        // TODO
+        let fmt = Format::new(640, 480, FourCC::new(launch_vid));
```
QuantumEntangledAndy, line 122: If you listen to the `BcMediaInfoV2` and `BcMediaInfoV1` packets as suggested above you can set the sizes using that

QuantumEntangledAndy (commented):
I've made some general comments if you could tell more about your specific issues I could give you some more specific comments

QuantumEntangledAndy:
We are probably also going to want a bunch of `#[cfg(target_os = "linux"]] ` so that this code is only avaliable on linux. Read up on how to do that [here](https://doc.rust-lang.org/reference/conditional-compilation.html)

0ry0n:
Ok, I have corrected these problems.
Now the problem is that I don't really know how to properly store `device` and `stream` in `struct V4ltOutputs`.
I relied on the example given in the [libv4l](https://github.com/raymanfx/libv4l-rs/blob/master/examples/stream_forward_mmap.rs) module where `device` and `stream` are `& mut`.
I tried to implement them this way but the compiler gives me lifecicle problems and I have no idea how to fix them.

Thanks for your help!

src/v4l/mod.rs
```diff
@@ -0,0 +1,248 @@
...
+            let arc_cam = Arc::new(camera);
+
+            if let Some(v4l_device) = arc_cam.v4l_device {
+                if ["mainStream"].iter().any(|&e| e == arc_cam.v4l_stream) {
+                    let v4l = V4lDevice::new(v4l_device as usize);
+                    let mut outputs = v4l
+                        .add_stream()
+                        .unwrap();
+                    let main_camera = arc_cam.clone();
+                    s.spawn(move |_| camera_loop(&*main_camera, Stream::Main, &mut outputs, true));
```
QuantumEntangledAndy, line 62: Remove your `&*` this will deference then reference the variable (then your compiler will deterrence it again because the type of the function sig is `Device` not `&Device`. Also try to give the `stream` without an `&` too. So that your structure can take ownership but I'm not sure how `v4l.add_stream` works. Check the doc to see if it returns a reference or not. If it returns a reference we have to work out something different.

src/v4l/v4lt.rs
```diff
@@ -0,0 +1,166 @@
...
+
+type Result<T> = std::result::Result<T, ()>;
+
+pub(crate) struct V4lDevice {
+    device: Device,
+}
+
+pub(crate) struct V4ltOutputs<'a> {
+    device: Device,
+    stream: &'a mut MmapStream<'a>,
```
QuantumEntangledAndy, line 19: First I think we need to remove the `&mut` at the front. You seem to have ownership of it at this point. 

However I suspect that there might be another issue in that the `MmapStream` has a lifetime parameter. This usually means that there is a reference in it to that points to data else where. The compiler needs to be certain that the lifetime of `'a` does not die. Which it cannot be certain of because of it running on a different thread. This will be an issue.

QuantumEntangledAndy:
Have a look at the PR I made against your branch. https://github.com/0ry0n/neolink/pull/1

src/v4l/v4lt.rs
```diff
@@ -0,0 +1,184 @@
...
+            video_format: None,
+        };
+        Ok(result)
+    }
+
+    pub(crate) fn run(&mut self) -> Result<()> {
+        // After we have created the device stream we cannot
+        // edit the height/width etc
+        // So first we pull packets from the camera until we have
+        // enough data to setup the height etc
```
thirtythreeforty, line 71: Is there a way you can query metadata without parsing the video itself? There might be a camera Xml you can fetch that will tell you what its current settings are. (I've seen something similar but I can't recall if it's _available_ or _active_ settings.)

src/v4l/mod.rs
```diff
@@ -0,0 +1,270 @@
+///
+/// # Neolink V4l
```
thirtythreeforty, line 2: This file appears to be about 95% copy-paste from the default stream command. The camera drive and setup loops are the same. Can we break those out into a common module and pass a "sink" trait to that module?  I think the Gstreamer module is already kinda structured that way, so it would be an easy change to the existing code.

src/v4l/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+///
+/// This module serves the v4l streams for the
+/// `neolink v4l` subcommand
```
thirtythreeforty, line 5: I object slightly to `v4l` as a name. Presumably Windows users would like something like this too and a hypothetical implementation using native Windows APIs would not be named `v4l`. What's a good name for the _feature_, not the mechanism?

src/config.rs
```diff
@@ -73,6 +79,26 @@ pub(crate) struct CameraConfig {
     #[validate(range(min = 0, max = 31, message = "Invalid channel", code = "channel_id"))]
     #[serde(default = "default_channel_id")]
     pub(crate) channel_id: u8,
+
+    #[cfg(target_os = "linux")]
+    // Maximum number of devices allowed by v4l2loopback (8)
```
thirtythreeforty, line 84: It doesn't look like this limit is actually a hard limit, since it's possible to create additional devices:

https://github.com/umlaeute/v4l2loopback/issues/367

Can you move this check to runtime, and fail with a nice error message if the user tries to create a device with a number that the kernel turns out not to like?

thirtythreeforty (changes requested):
Very interesting feature! Thank you for implementing it and thanks @QuantumEntangledAndy for the initial code review. I hadn't even considered doing something like this but I think it is a good idea.

A few points of feedback. I think the largest one is the code duplication in the wrapper module, and perhaps also (bikeshedding?) the name.

thirtythreeforty:
And one last thing, don't forget to add the command to the readme under "additional commands"!

src/v4l/mod.rs
QuantumEntangledAndy: Just as a reference: Macos has CoreMediaIO and windows has DirectShow for virtual camera. Neither of which seem to have any high level crates, it's just an sys level FFI stuff at the moment, we'd have to roll it ourselves.

We could perhaps get similar functionality though with just a fifo exposing the current streams.

I do kinda agree that the name should be generic rather than `v4l` (video 4 linux). Maybe `virtual_camera`?

src/v4l/v4lt.rs
QuantumEntangledAndy: This data is also available in
- `<StreamInfoList>` xml (msg id: 146)
- `<Compression>` xml (msg id: 56)
- Initial login (msg id 1)

But this MediaInfo packet was the easiest as the `StreamInfoList` and `Compression` aren't currently programmed in

src/v4l/mod.rs
0ry0n: `virtcam`?

PR #261, opened 2022-03-26
use new command 'neolink rtsp' in docs

sebix:
on some places in the docs the old command 'neolink ...' was used.
that raises a warning, so refer to 'neolink rtsp' where apropriate


QuantumEntangledAndy (approved):

QuantumEntangledAndy:
Thanks.

PR #271, opened 2022-05-09
Add new motion command to set PIR sensor ON or OFF on battery powered camera

jbremond:
tested on reolink argus pro


QuantumEntangledAndy:
This all looks ok the only concern I have is with the choice of name for motion. As it would seem to clash with the other sort of non-pir motion detection. 

Also do you know if having this on will change the messages received from the motion alarm messages as used in #203. The motion alarm messages include things like `MD` for motion detected using frame difference and I wonder if it instead reports something different when PIR is on.

jbremond:
thanks @QuantumEntangledAndy for your reply
Regarding your first question:
> This all looks ok the only concern I have is with the choice of name for motion. As it would seem to clash with the other sort of non-pir motion detection.
>
You are rigth, I can rename the feature, variables, and commande to PIR instead of motion, what do you think ?

> Also do you know if having this on will change the messages received from the motion alarm messages as used in #203. The motion alarm messages include things like `MD` for motion detected using frame difference and I wonder if it instead reports something different when PIR is on.

I don't think there is an impact on that message

QuantumEntangledAndy:
Thanks, I think to pir would be good to keep them separate.

jbremond:
I just put a new commit to do the renaming motion --> PIR

QuantumEntangledAndy:
Could you `rust fmt` this. Just run

```bash
cargo fmt
```

in the root dir

jbremond:
I just pushed the result of the formatted code

QuantumEntangledAndy:
Thanks once all the checks run ill merge it. Do you have any other plans for neolink?

QuantumEntangledAndy (approved):

PR #282, opened 2022-05-30
Fix typo in help for talk command

cincodenada:
Thanks for this project - got my Reolink that I picked up on a whim at the thrift store up and running like a charm!

Noticed a couple typos when digging around the help, so I figured I'd fix 'em - I assume these triple-slash comments are somehow generated into documentation (which is super rad, taking notes for my next Rust project!)

### DIFF
--- a/src/talk/cmdline.rs
+++ b/src/talk/cmdline.rs
@@ -9,13 +9,13 @@ use structopt::StructOpt;
 /// `gst-launch` can be used to prepare this data
 #[derive(StructOpt, Debug)]
 pub struct Opt {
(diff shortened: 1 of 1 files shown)


QuantumEntangledAndy:
Glad to be of help. We are a bit understaffed here at the moment, so if you fancy helping with anything just open another PR. Rust is quite cool once you wrap your head around the borrow checker, and if you need help with anything just give us a shout.

Yeah triple slashes will become manuals for function/strucs etc while `\\!` is module (file) level doc coments.

thirtythreeforty (approved):

thirtythreeforty:
Thanks for the close eyes!

PR #330, opened 2022-12-13
Rework streams

QuantumEntangledAndy:
Major changes to how streams work

Rtsp
----

- Using a state machine for the rtsp camera

Mqtt
----
- Changes to how motion is detected

Core
---
- Changes to stream and motion that return a struct that will manage and shutdown the connection automatically as well as allowing for abstractions on how we view the data
- Changes to connection so that it uses an `Arc`, this is in preperation for some changes to the threading model we employ


QuantumEntangledAndy:
Opps opened this in the wrong place

PR #342, opened 2023-02-24
Bugfix/aes login

QuantumEntangledAndy:
Fixes issue with using AES too early during login


QuantumEntangledAndy:
And again.... Sorry, github always defaults to the parent fork

PR #350, opened 2023-04-04
Add command to read PIR status

jl1990:
This PR adds a command for reading PIR status.

### DIFF
--- a/docs/unix_setup.md
+++ b/docs/unix_setup.md
@@ -211,7 +211,7 @@ If the mainStream fails but the subStream succeeds you may need to add
 
 ## Common Errors
 
-- Missing dependancies
+- Missing dependencies
 
   If when start neolink you get a message such as:
 
--- a/docs/unix_setup.md
+++ b/docs/unix_setup.md
@@ -9,7 +9,7 @@ correct names yourselves.
 
 The general steps we will follow in this guide are as follows:
 
-1. Install dependancies
+1. Install dependencies
 
 2. Download neolink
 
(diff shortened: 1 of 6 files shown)


QuantumEntangledAndy:
Are you aware of my fork which has this?

We have discovered some issues where for battery cameras it will return error code 400 intermittently when trying to fetch the PIR data. We think this happens if the battery camera is not given enough time to wake up from some sort of sleep mode.

QuantumEntangledAndy:
This original fork here hasn't been actively maintained for a few years now. Just been me doing little maintenance things. With what merge permissions I have. 

As a  result I have move active delvelopment onto my fork.

jl1990:
Hi @QuantumEntangledAndy, I was not aware of your fork, but I am very happy to see there is a more updated branch.

I will have a look. Thanks for warning!

QuantumEntangledAndy:
Sure PIR status just for merged today and should be out in 0.5.6.

Discussions about the feature and the driving force for the updates are here https://github.com/QuantumEntangledAndy/neolink/issues/51

While the actual PR that adds the updated PIR is here https://github.com/QuantumEntangledAndy/neolink/pull/53

jl1990:
It's great, I was so surprised to see you also included MQTT integration. I was using this code for integrating with my HA and your fork is perfect for that

QuantumEntangledAndy:
Hope its helpful.

I don't have any testers for the mqtt so if you can open any issues that you find it would be good

PR #356, opened 2023-05-13
Add PTZ control via direct movement and presets

danielkaiser:
This PR adds the ability to control the cameras PTZ movement.

The new command line functions can either perform movement for a given time in a given direction or read out, set and move to the cameras position presets.

As I only have a Reolink E1 for development and the camera does not feature a zoom function there are no commands/options to zoom yet.

### DIFF
--- a/crates/core/src/bc/model.rs
+++ b/crates/core/src/bc/model.rs
@@ -14,6 +14,10 @@ pub const MSG_ID_VIDEO: u32 = 3;
 pub const MSG_ID_TALKABILITY: u32 = 10;
 /// TalkReset messages have this ID
 pub const MSG_ID_TALKRESET: u32 = 11;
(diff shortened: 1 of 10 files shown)


QuantumEntangledAndy:
Welcome, this mostly seems good and I can merge it here if you want.

But I should point out that this fork has not really had any active development for quite some time. Most active development is happening [here](https://github.com/QuantumEntangledAndy/neolink), we have some of this PTZ there integrated into the MQTT interface (but not the preset stuff)

danielkaiser:
Thanks for the info. I rebased my changes to the other fork and opened https://github.com/QuantumEntangledAndy/neolink/pull/92 there. 

Maybe a note about the more active development fork in the main repo would be helpful.

PR #369, opened 2023-08-03
Fix/async lookup and register

QuantumEntangledAndy:
This attempts to change the way in which reolink registration servers as contacted.

Previously we would stop at the first sucessful reply from one of the reolink servers, but it seems that sometimes those servers had outdated information.

Now we keep waiting for more replies (while trying current replies) and only stop when either all replies fail or full complete and valid registration data is obtained.

We also now only register once rather once per registration method


QuantumEntangledAndy:
Sorry default branch on github is always the fork source....

PR #376, opened 2024-04-05
Updates to deps

QuantumEntangledAndy:
Mostly an update to dependencies

Couple of dependencies needed updating for newer rust so swapped them out
- Swapped yaserde for the usual serde and quick_xml
- Swapped err_derive for thiserror

Also the push notifications will stop working in a few months (google are changing their APIs), the fcm package has an update ready but I need additional data (reolinks pub key) this should be in the appdata (assuming reolink have updated their app) but I couldn't find it. Maybe they are waiting until we closer to googles update deadline?

Also trying out using shared gst pipelines.


QuantumEntangledAndy:
Sorry it defaults to the root repo everytime

PR #377, opened 2024-04-11
Fix CI

QuantumEntangledAndy:
Should fix the failing CI


QuantumEntangledAndy:
Everything except docker seems to work (clippy has a lot of warning too but I'm not fixing them here)

thirtythreeforty (approved):

thirtythreeforty:
Thank you for this!

